PR #26, opened 2021-06-01
LiveQuery does not match queries correctly

### LINKED ISSUE #23: `LiveQuery` does not match queries correctly
It shall compare query `_class` as well

BykhovDenis:
Fixes #23

### DIFF
--- a/packages/core/src/storage.ts
+++ b/packages/core/src/storage.ts
@@ -23,7 +23,7 @@ export type DocumentQuery<T extends Doc> = Partial<Data<T>>
 export interface Storage {
   findAll<T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>): Promise<T[]>
   tx(tx: Tx): Promise<void>
-  isDerived(_class: Ref<Class<Obj>>, _extendedClass: Ref<Class<Obj>>): boolean
+  isDerived<T extends Obj>(_class: Ref<Class<T>>, from: Ref<Class<T>>): boolean
 }
 
 export class TxProcessor {
--- a/packages/core/src/storage.ts
+++ b/packages/core/src/storage.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 //
 
-import type { Emb, Doc, Ref, Class, Data } from './classes'
+import type { Emb, Doc, Ref, Class, Data, Obj } from './classes'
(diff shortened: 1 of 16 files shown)


packages/query/src/index.ts
```diff
@@ -26,10 +26,11 @@ type Query = {
   callback: (result: Doc[]) => void
 }
 
-function match(query: DocumentQuery<Doc>, attributes: Data<Doc>): boolean {
-  for (const key in query) {
-    const value = (query as any)[key]
-    if ((attributes as any)[key] !== value)
+function match(q: Query, tx: TxCreateObject<Doc>): boolean {
+  if (q._class !== tx.objectClass) return false
```
z2sx, line 30: This is not correct! Classes can be derived, so consider `Person` <- `Student` hierarchy for example. Query can be for all `Persons` having `Вася` name. Then new `Student` object named `Вася` comes in... So this tx will match query.

z2sx (changes requested):

z2sx:
Also, please take jest.config.js from platform to fix CI coverage reporter.

packages/query/src/index.ts
BykhovDenis: Ok. Can I add method in Storage for use hierarchy.getAncestors for check query class in tx objectClass's ancestors?

packages/core/src/storage.ts
```diff
@@ -23,6 +23,7 @@ export type DocumentQuery<T extends Doc> = Partial<Data<T>>
 export interface Storage {
   findAll<T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>): Promise<T[]>
   tx(tx: Tx): Promise<void>
+  isDerived(_class: Ref<Class<Obj>>, _extendedClass: Ref<Class<Obj>>): boolean
```
z2sx, line 26: I would love to have this parametrized by `<T extends Obj>`, so at least users can not supply "is Obj extends Person".

Another thing is I do not understand parameters meaning: if I want to check that 'Student extends Person" should I `isDerived(Person, Student)` or `isDerived(Student, Person`? I suggest to rename second parameter to `from`, so it will be clear `isDerived(Student, Person) === true`

z2sx (changes requested):

z2sx (approved):

PR #28, opened 2021-06-02
Fixed get translation when incorrect string

### LINKED ISSUE #4: Invalid `label` for `EditBox` result in Unhandled Promise Rejection
Just try `<EditBox label="Search"/>`

BykhovDenis:
Fixes #4 

Signed-off-by: Denis Bykhov <bykhov.denis@gmail.com>

### DIFF
--- a/packages/platform/src/i18n.ts
+++ b/packages/platform/src/i18n.ts
@@ -52,6 +52,7 @@ async function loadTranslationsForComponent (component: Component): Promise<Mess
 }
 
 async function getTranslation (message: IntlString): Promise<IntlString | Status> {
+  if (message.indexOf('.') === -1) return message
   const id = parseId(message)
   let messages = translations.get(id.component)
   if (messages === undefined) {
--- a/packages/platform/src/i18n.ts
+++ b/packages/platform/src/i18n.ts
@@ -52,17 +52,21 @@ async function loadTranslationsForComponent (component: Component): Promise<Mess
 }
 
 async function getTranslation (message: IntlString): Promise<IntlString | Status> {
-  if (message.indexOf('.') === -1) return message
-  const id = parseId(message)
-  let messages = translations.get(id.component)
-  if (messages === undefined) {
-    messages = await loadTranslationsForComponent(id.component)
-    translations.set(id.component, messages)
+  try {
+    const id = parseId(message)
+    let messages = translations.get(id.component)
+    if (messages === undefined) {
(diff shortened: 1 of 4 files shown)


packages/platform/src/i18n.ts
```diff
@@ -52,6 +52,7 @@ async function loadTranslationsForComponent (component: Component): Promise<Mess
 }
 
 async function getTranslation (message: IntlString): Promise<IntlString | Status> {
+  if (message.indexOf('.') === -1) return message
```
z2sx, line 55: This is not correct. A message can contain '.' for example 'Привет.'. We should verify if translation provided and fallback to original `message` if it is not.

z2sx (changes requested):

z2sx:
Also, please provide a test case for this. I guess we do not have one since all tests were passed well.

packages/platform/src/i18n.ts
BykhovDenis: Just use try/catch when `getTranslation` call? I'm not sure if this is good, perhaps it would be better if `parseId` would not throw an error, but return `undefined`. What's the general concept about errors?

z2sx: Oh got it, so the problem is that `parseId` throws exception... so then your fix makes sense. But let's create `validateId` together with `parseId` and use it instead of `message.indexOf('.') === -1`. `validateId` implementation should match `parseId` so if `parseId` results in exception, then `validateId` returns `false`.

packages/platform/src/i18n.ts
```diff
@@ -52,16 +52,21 @@ async function loadTranslationsForComponent (component: Component): Promise<Mess
...
+      messages = await loadTranslationsForComponent(id.component)
+      translations.set(id.component, messages)
+    }
+    if (messages instanceof Status) {
+      return messages
+    }
+    return (id.kind ? (messages[id.kind] as Record<string, IntlString>)?.[id.name] : messages[id.name] as IntlString) ?? message
+  } catch (err) {
+    setPlatformStatus(err)
+    return err
```
z2sx, line 68: This is not what we want, we want original string to be returned (but only in case when parseId failed). I propose to get back to your original version, ideally introducing `isValidId`.

z2sx (changes requested):

z2sx:
Fixes #4

z2sx (approved):

PR #29, opened 2021-06-02
implement theme switcher

BykhovDenis:


packages/ui/src/components/internal/ThemeSelector.svelte
```diff
@@ -1,28 +1,34 @@
...
+        current = Themes.GREY
+        break;
+      case Themes.GREY:
+        current = Themes.DARK
+        break;
+      case Themes.DARK:
+        current = Themes.LIGHT
+        break;
+    }
+    setTheme(current)
```
z2sx, line 20: This looks to be overkill :) why not `setTheme(themes[current % themes.length])` where `themes = ['light', 'dark',...]`

z2sx (changes requested):

packages/ui/src/components/internal/ThemeSelector.svelte
BykhovDenis: Some overengineering, I think in future we need changing icons (when we will have them), and store user setting

z2sx: So? :)

z2sx: Ping @BykhovDenis

z2sx (approved):

PR #34, opened 2021-06-03
Cache in livequery

BykhovDenis:


packages/query/src/index.ts
```diff
@@ -53,24 +53,44 @@ export class LiveQuery extends TxProcessor implements Storage {
...
-    return this.storage.findAll(query._class, query.query).then(result => query.callback(result))
+  private updateCache<T extends Doc>(query: Query, object: T): void {
+    const values = this.cache.get(query) || []
+    const index = values.findIndex((doc) => doc._id === object._id)
+    if (index === -1) {
+      values.push(object)
+    } else {
+      values[index] = object
+    }
+    this.cache.set(query, values)
```
z2sx, line 64: you do not need to do this, `values` remains same and still in cache

packages/query/src/index.ts
```diff
@@ -53,24 +53,44 @@ export class LiveQuery extends TxProcessor implements Storage {
     return true
   }
 
-  private refresh(query: Query): Promise<void> {
-    return this.storage.findAll(query._class, query.query).then(result => query.callback(result))
+  private updateCache<T extends Doc>(query: Query, object: T): void {
```
z2sx, line 56: We will have various update logic for different transaction types, so I suggest to rename this to reflect "create object update", such as `updateCreateObject`

z2sx (changes requested):

packages/query/src/index.ts
BykhovDenis: I suggest `createObjectCaching` like transaction type + 'Caching'? Because for update transaction `updateUpdateObject` so strange)

z2sx: I would be more happy with common prefix, not suffix :) maybe `cacheXXX` like `cacheCreateObject`

z2sx (approved):

PR #55, opened 2021-06-06
Update RequestProcessor to not depend on Core

### LINKED ISSUE #44: Move `RequestProcessor` out of `rpc` and remove `core` dependency
`rpc` package defines JsonRPC protocol, and used from other places such as `login-impl`. Some applications such as `login` does not need `core` but need `rpc`.

With `RequestProcessor` dependency to `core` being introduced, which results in full core to be compiled into `login` app bundle.

I suggest to use `packages/client` package for this stuff, which already depends on `core`.

haiodo:
Closes #44


packages/rpc/src/index.ts
```diff
@@ -13,11 +13,9 @@
 // limitations under the License.
 //
 
-import { Class, Doc, DocumentQuery, Ref, Storage, Tx } from '@anticrm/core'
 import { component, Component, PlatformError, Severity, Status, StatusCode } from '@anticrm/status'
 
-export type ReqId = string | number
-
+export type ReqId = string
```
z2sx, line 18: Why `ReqId` is not a `string | number` anymore? -- this is according to JsonRPC spec.

packages/core/src/request.ts
```diff
@@ -0,0 +1,96 @@
...
+  requests = new Map<string, Operation>()
+  stream: () => Promise<RequestStream>
+  handler: (tx: Tx) => void
+
+  constructor (stream: () => Promise<RequestStream>, handler: (tx: Tx) => void) {
+    this.stream = stream
+    this.handler = handler
+  }
+
+  public process(id: string | undefined, result: any, error?: any): void {
```
z2sx, line 34: request id should be `ReqId` type

z2sx (changes requested):

packages/rpc/src/index.ts
haiodo: Sure, let me make it according to spec.

packages/core/src/request.ts
haiodo: Fixed.

packages/core/src/request.ts
```diff
@@ -0,0 +1,100 @@
+import { Status } from '@anticrm/status'
+import { Class, Doc, Ref } from './classes'
+import { DocumentQuery, Storage } from './storage'
+import { Tx } from './tx'
+
+export type ReqId = string | number
+
+export interface RequestStream {
+  send: (id: ReqId, method: string, params: any[]) => void
```
z2sx, line 9: send should accept `Request<T>` from `rpc`

z2sx (changes requested):

packages/core/src/request.ts
```diff
@@ -0,0 +1,96 @@
...
+export interface RequestStream {
+  send: (id: string, method: string, params: any[]) => void
+}
+
+interface Operation {
+  id: string
+  request: () => void // Perform request again
+  promise: Promise<any>
+  resolve: <T>(value: T) => void
+  reject: (reason?: any) => void
```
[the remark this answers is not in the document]
haiodo: It does, but there is no Deferred promise concept, so this is only way to complete it from outside.

packages/core/src/request.ts
haiodo: Fixed.

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
 
 export interface RequestStream {
-  send: <T extends any[]>(req: Request<T>) => void
+  send: (request: Request<any>) => void
 }
 
 interface Operation {
-  id: number
-  request: () => void // Perform request again
+  id: ReqId
```
z2sx, line 73: Why do we need this?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
   promise: Promise<any>
-  resolve: <T> (value: T) => void
+  resolve: <T>(value: T) => void
```
z2sx, line 75: Why do we need <T>?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
+  reject: (reason?: any) => void
```
z2sx, line 76: Is not it `reason: error`?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
+   * @param error - error to reject with.
+   */
+  public reject (error: any): void {
     // We need to reply requests in case they are missed.
     for (const op of this.requests.entries()) {
-      op[1].request()
+      op[1].reject(error)
     }
+    this.requests.clear()
   }
```
z2sx, line 117: Why do we need this method?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
   public async request (method: string, ...params: any[]): Promise<any> {
-    const id = ++this.reqId
-    let resolve: <T> (value: T) => void = () => {}
+    const id = ++this.reqIndex
+    let resolve: <T>(value: T) => void = () => {}
+    let reject: (reason?: any) => void = () => {}
 
-    const p = new Promise((r) => { resolve = r }) // eslint-disable-line
+    // eslint-disable-next-line promise/param-names
+    const p = new Promise((r, rej) => {
```
z2sx, line 125: can we name parameters `res`, `rej`?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
+  public reject (error: any): void {
     // We need to reply requests in case they are missed.
     for (const op of this.requests.entries()) {
-      op[1].request()
+      op[1].reject(error)
     }
+    this.requests.clear()
   }
 
   public async request (method: string, ...params: any[]): Promise<any> {
```
z2sx, line 119: Is it possible to change params to `Request<T>`?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
-export class RequestProcessor implements Storage {
-  reqId: number = 0
+export class RequestProcessor {
+  reqIndex: number = 0
   requests = new Map<ReqId, Operation>()
   stream: () => Promise<RequestStream>
-  handler: (tx: Tx) => void
 
-  constructor (stream: () => Promise<RequestStream>, handler: (tx: Tx) => void) {
+  constructor (stream: () => Promise<RequestStream>) {
```
z2sx, line 90: Can we have `stream` to be `RequestStream`? And if we can, I believe using `(request: Request<any>) => void` is better than RequestStream -- no need to look what `RequestStream` means. Also, why not `(request: Request<any>) => Promise<void>`?

plugins/plugin-core-impl/src/connection.ts
```diff
@@ -32,18 +32,25 @@ export async function connect (
...
-            processor.onOpen()
           }
-          ws.onerror = () => {
+          ws.onerror = (reason) => {
+            // Reject all pending operations.
+            processor.reject(reason)
+
             // Force reconnect in 1 second.
-            setTimeout(() => { doConnect() }, 1000) // eslint-disable-line
+            setTimeout(async () => await doConnect(), 1000) // eslint-disable-line @typescript-eslint/no-misused-promises
```
z2sx, line 44: Let's remove `reconnect` and error handling stuff. Until we have scenarios how to test this.

plugins/plugin-core-impl/src/connection.ts
```diff
@@ -53,15 +60,21 @@ export async function connect (
...
   const processor = new RequestProcessor(async () => {
     const s = await doConnect()
-    return {
-      send: (request) => {
-        s.send(serialize(request))
-      }
-    }
-  }, handler)
+    return { send: (req) => s.send(serialize(req)) }
+  })
```
z2sx, line 64: I do not get this complication, why not `new RequestProcessor((req) => s.send(serialize(req)))`?

plugins/plugin-core-impl/src/connection.ts
```diff
@@ -32,18 +32,25 @@ export async function connect (
...
+            if (resp.id !== undefined) {
+              processor.process(resp)
+            } else if (resp.result !== undefined) {
+              handler(resp.result as Tx)
+            }
           }
-          ws.onmessage = evt => { processor.process(readResponse(evt.data as string)) }
         })
       }
     } catch (err) {
```
z2sx, line 56: There is nothing within `catch` block. Is it intentionally?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,83 @@ export const Code = component('rpc' as Component, {
...
     this.stream = stream
-    this.handler = handler
   }
 
-  public process<T>(response: Response<T>): void {
+  public process (response: Response<any>): void {
```
z2sx, line 94: ideally this shall be `protected` -- need to inherit "connection" from RequestProcessor in this case

plugins/plugin-core-impl/src/connection.ts
```diff
@@ -53,15 +60,21 @@ export async function connect (
...
+    async findAll<T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>): Promise<T[]> {
+      const result = await processor.request('findAll', _class, query)
+      return result as T[]
+    },
+    async tx (tx: Tx): Promise<void> {
+      await processor.request('tx', tx)
+      // Process on server and return result.
+      handler(tx)
+    }
+  }
```
z2sx, line 79: Please think to inherit `connection` from `RequestProcessor` and implement `Storage` -- will put many things in shape in this case.

z2sx (changes requested):

packages/rpc/src/index.ts
haiodo: Nope, 
Screenshot 2021-06-09 at 11 00 51

packages/rpc/src/index.ts
haiodo: Fixed.

packages/rpc/src/index.ts
haiodo: Removed.

packages/rpc/src/index.ts
haiodo: To not hang on promise in UI in case of connection drop.

plugins/plugin-core-impl/src/connection.ts
haiodo: Done.

plugins/plugin-core-impl/src/connection.ts
haiodo: Refactored.

packages/rpc/src/index.ts
haiodo: Done.

plugins/plugin-core-impl/src/connection.ts
haiodo: Done.

plugins/plugin-core-impl/src/connection.ts
haiodo: Done.

packages/rpc/src/index.ts
z2sx: Can you please elaborate on this more?

packages/rpc/src/index.ts
```diff
@@ -62,89 +62,75 @@ export const Code = component('rpc' as Component, {
...
-    const result = await this.request('findAll', _class, query)
-    return result as T[]
-  }
-
-  async tx (tx: Tx): Promise<void> {
-    await this.request('tx', tx)
-    // Process on server and return result.
-    this.handler(tx)
+    return await promise.promise.catch((reason) => {
+      throw new PlatformError(new Status(Severity.ERROR, Code.ConnectionError, { message: JSON.stringify(reason ?? '') }))
```
z2sx, line 133: Is it really "Connection Error" or something else can happen? I'd prefer to use `unknownErr()` here to pass potential PlatformError forward.

z2sx (changes requested):

plugins/plugin-core-impl/src/connection.ts
```diff
@@ -12,56 +12,49 @@
...
-          const ws = new WebSocket(`ws://${clientUrl}/${clientId}`)
-          ws.onopen = () => {
-            resolve(ws)
+  class WebSocketConnection extends RequestProcessor implements Storage {
+    constructor () {
+      super()
+      socket.onerror = (reason) => this.reject('Connection error')
+      socket.onclose = (reason) => this.reject('Connection closed')
+      socket.onmessage = (evt) => this.onmessage(evt)
+    }
```
z2sx, line 33: I think making `socket` and `handler` private fields and moving class declaration out of `connect` would be better to understand it's logic.

z2sx (changes requested):
`RequestProcessor` and `Connection` becomes clear, let's do couple of changes more... What really needs refactoring is server, which is a bit of mess, let me suggest some fixes in the next review.

haiodo:
Server will be updated on next steps, with security work. Let me process comments :-)

packages/rpc/src/index.ts
haiodo: Updated.

plugins/plugin-core-impl/src/connection.ts
haiodo: Moved, Not sure if it make it any more clear :-)

z2sx (approved):

PR #56, opened 2021-06-07
implement `storage` for `elasticsearch`

### LINKED ISSUE #41: Implement `Storage` for `ElasticSearch`

BykhovDenis:
Closes #41


packages/core/src/esdb.ts
```diff
@@ -0,0 +1,110 @@
...
+import { TxProcessor } from './storage'
+import type { Emb, Doc, Ref, Class, Data, Collection } from './classes'
+import type { TxAddCollection, TxCreateObject } from './tx'
+import type { Hierarchy } from './hierarchy'
+import { Client, RequestParams } from '@elastic/elasticsearch'
+import { generateId } from './utils'
+
+export class EsDb extends TxProcessor implements Storage {
+  private readonly hierarchy: Hierarchy
+  private readonly client = new Client({
```
haiodo, line 26: I propose to client connection details to EsDB and rename it to ElasticStorage. 
Assume we potentially could use different Elastic storages for different domains.

packages/core/src/esdb.ts
```diff
@@ -0,0 +1,110 @@
...
+    node: process.env.ES_DB_URL ?? 'http://localhost:9200',
+    auth: {
+      username: process.env.ES_DB_USERNAME ?? 'elastic',
+      password: process.env.ES_DB_PASSWORD ?? 'changeme'
+    }
+  })
+
+
+  constructor (hierarchy: Hierarchy) {
```
haiodo, line 35: Please accept a workspaceId, to allow few workspaces to be hold on same server.

packages/core/src/esdb.ts
```diff
@@ -0,0 +1,110 @@
...
+    const { body } = await this.client.get(request)
+    
+    return {_id: body._id, ...body._source}
+  }
+  
+  async findAll<T extends Doc>(_class: Ref<Class<T>>, query: Partial<Data<T>> & Partial<Doc>): Promise<T[]> {
+    const result:T[] = []
+
+    const criteries = []
+    for (const key in query) {
```
haiodo, line 54: You should handle a `_class` field and extract a full list of classes extended this one passed in search to find all possible documents.

packages/core/src/esdb.ts
```diff
@@ -0,0 +1,110 @@
...
+    await this.client.index(object)
+  }
+
+  protected async txAddCollection (tx: TxAddCollection<Emb>): Promise<void> {
+    const doc = await this.objectById(tx.objectId)
+    if ((doc as any)[tx.collection] === undefined) {
+      (doc as any)[tx.collection] = {} as Collection<Emb>
+    }
+    const collection = (doc as any)[tx.collection]
+    collection[tx.localId ?? generateId()] = tx.attributes
```
haiodo, line 99: Should we somehow check if duplicate localId is passed?

packages/core/src/esdb.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return result as T[]
+  }
+
+  protected async txCreateObject (tx: TxCreateObject<Doc>): Promise<void> {
+    const object: RequestParams.Index = {
+      id: tx.objectId,
+      index: 'workspace',
+      type: tx.domain,
+      body: { _class: tx.objectClass, ...tx.attributes}
```
haiodo, line 88: Please use hierarchy.getClass() to store a class for Mixins.

packages/core/src/esdb.ts
BykhovDenis: I think it may be useful for update

packages/core/src/elasticStorage.ts
```diff
@@ -0,0 +1,131 @@
...
+  username: string,
+  password: string
+}
+
+export class ElasticStorage extends TxProcessor implements Storage {
+  private readonly hierarchy: Hierarchy
+  private readonly client: Client
+  private readonly workspace: string
+
+  constructor (hierarchy: Hierarchy, workspace: string, connection: ConnectionParams) {
```
haiodo, line 35:
```suggestion
  constructor (hierarchy: Hierarchy, connection: ConnectionParams, workspace?: string) {
```

packages/core/src/elasticStorage.ts
```diff
@@ -0,0 +1,131 @@
...
+    super()
+    this.hierarchy = hierarchy
+    this.workspace = workspace ?? 'workspace'
```
haiodo, line 38: It is not clear what workspace param passed could be undefined, please use ? or explicitly specify type | undefined in definition.

packages/core/src/hierarchy.ts
```diff
@@ -21,6 +21,7 @@ import core from './component'
 export class Hierarchy {
 
   private readonly classes = new Map<Ref<Class<Obj>>, Data<Class<Obj>>>()
+  private readonly extends = new Map<Ref<Class<Obj>>, Array<Ref<Class<Obj>>>>()
```
haiodo, line 24: It would be better to push this changes as separate PR.

z2sx (changes requested):
From the bird flight view it looks good, but the `core` is not the place where we want this -- at least we do not want `core` to be dependent on `elastic` packages. `server/elastic` is a good place for this stuff.

server/elastic/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    })
+
+    for (const doc of body.hits.hits) {
+      result.push({ _id: doc._id, ...doc._source })
+    }
+
+    return result
+  }
+
+  protected async TxCreateDoc (tx: TxCreateDoc<Doc>): Promise<void> {
```
z2sx, line 100: this will not work -- use lowerCase

z2sx: Please consider writing test for this

z2sx (changes requested):

packages/core/src/hierarchy.ts
```diff
@@ -70,4 +72,20 @@ export class Hierarchy {
...
+    return this.extends.get(_class) ?? []
+  }
+
+  private addExtends<T extends Obj>(_class: Ref<Class<T>>): void {
+    const hierarchy = this.getAncestors(_class as Ref<Class<Obj>>)
+    for (const cls of hierarchy) {
+      const list = this.extends.get(cls)
+      if (list === undefined) {
+        this.extends.set(cls, [_class as Ref<Class<Obj>>])
+      } else if (!list.includes(_class)) {
```
z2sx, line 86: `list` will never include a class, is not it?

z2sx (changes requested):

packages/core/src/hierarchy.ts
BykhovDenis: Never in normal case only, but if the transaction is applied twice, or the objectId is repeated in different transactions, it will cause duplicate data

z2sx: Create transaction can't happen twice

z2sx: and objectId with create transaction can't appear twice, if you want to preserve this check I would better recommend to throw exception if there is class in the list, but for performance reasons I'd prefer to remove this check

z2sx (approved):

PR #76, opened 2021-06-09
Unify VDoc ops into Ops

### LINKED ISSUE #75: Refactor `Doc` and `VDoc` transactions
`user` and `timestamp` makes sense for `Tx` (not `VDoc`) transactions as well, so it looks like we can avoid `VDoc` transactions at all. Remaining question is `space` with is required for `VDoc` and makes no sense for `model` domain, but I think if we provide `undefined` space for model transactions everything will be fine. Who knows, later we may employ spaces for `model` domain as well.

So I propose to add following properties to `Tx` class:

* `space?: Ref<Space>`
* `user: Ref<Account>`
* `timestamp: number`

Very likely `user` and `timestamp` will be set on server side, but for now let's set them on client.

Thoughts?

haiodo:
1. Unify VDoc ops into Ops
2. Fix each-diff script.

Closes #75

### DIFF
--- a/common/changes/@anticrm/core/fix-75_2021-06-09-17-27.json
+++ b/common/changes/@anticrm/core/fix-75_2021-06-09-17-27.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@anticrm/core",
+      "comment": "Unify VDoc/Doc ops",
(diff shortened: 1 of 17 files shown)


z2sx (approved):
Looks super! Just what Doctor ordered! :)

PR #84, opened 2021-06-11
rewrite hierarchy's `getAncestors`

BykhovDenis:


packages/core/src/hierarchy.ts
```diff
@@ -92,4 +91,18 @@ export class Hierarchy {
...
+    let cl: Ref<Class<Obj>> | undefined = _class
+    while (cl !== undefined) {
+      const list = this.ancestors.get(_class)
+      if (list === undefined) {
+        this.ancestors.set(_class, [cl])
+      } else {
+        list.push(cl)
+      }
+      const attrs = this.classes.get(cl)
+      cl = attrs?.extends
```
haiodo, line 105: What if A extends B
And user add B and when A?

We should probably check all available descendants on moment of adding A?

BykhovDenis: If user try add class A extends class B when B is not exist will be throw error `const attrs = this.classes.get(cl)` here, and I think this is normal

BykhovDenis: And we can not check descendants because we use ancestors for create descendants

haiodo (approved):

haiodo:
Do we have any tests for ascendants?

z2sx (approved):

PR #86, opened 2021-06-11
Security describing specification

haiodo:
Security describing specification

(!) For discussion


haiodo:
Will not delete branch for now.

PR #94, opened 2021-06-14
Allow to reuse model.tx

haiodo:
1. Allow to reuse model.tx
2. Remove vdoc tests since they are not required amymore.


packages/core/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "rootDir": "./src",
     "strict": true,
     "sourceMap": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "esModuleInterop": true
```
z2sx, line 11: This shit did breaks things for me recently, so I removed it, what this is for?

z2sx (changes requested):

packages/core/tsconfig.json
z2sx: I mean `esModuleInterop`

z2sx (changes requested):

packages/core/tsconfig.json
haiodo: To allow just import not require, ok let me fix it as it was.

PR #96, opened 2021-06-14
Workbench application and space navigation

z2sx:


packages/core/src/memdb.ts
```diff
@@ -66,9 +66,13 @@ class MemDb {
 
   async findAll<T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>): Promise<T[]> {
     let result: Doc[]
-    if (query._id !== undefined) {
-      const obj = this.objectById.get(query._id)
-      result = obj !== undefined ? [obj] : []
+    if (Object.prototype.hasOwnProperty.call(query, '_id')) {
```
haiodo, line 69: Why query._id not work?

z2sx: do not work for queries like { _id: undefined }

z2sx: consider I have a query `{ _id: currentApp}`, so when `currentApp === undefined`, I think this query should return nothing. It was returning all the apps without this fix.

plugins/workbench-impl/src/components/navigator/SpacesNav.svelte
```diff
@@ -46,12 +47,19 @@
       showModal(model.createComponent, {})
     }
   }
+
+  function selectSpace(id: Ref<Space>) {
+    const loc = getCurrentLocation()
+    loc.path[2] = id
```
haiodo, line 53: I'm personally not like this approach, it is not flexible, how it work if not enought arguments are provided?

z2sx: I do not like this too, please suggest anything to make it good. Also I do not understand what you mean by "not enought arguments are provided"?

haiodo: How we could be sure we have at least 2 values in loc.path?

haiodo: Routes stuff you removed was my proposal :-)

z2sx: Let's discuss this on meeting, I tried to use routes stuff during weekend but it does not work well for me

packages/core/src/memdb.ts
haiodo: Ok.

PR #100, opened 2021-06-15
implement security storage

BykhovDenis:
Signed-off-by: Denis Bykhov <bykhov.denis@gmail.com>


server/server/src/securityStorage.ts
```diff
@@ -0,0 +1,89 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import core, { Class, Doc, Space, DocumentQuery, Ref, Storage, Tx, Account, TxCreateDoc, TxAddCollection, Emb, TxProcessor, makeEmb, Member, Hierarchy } from '@anticrm/core'
+
+export class SecurityStorage {
```
haiodo, line 18: I think we should separate into two stores, one fill fill and update model handling `tx` and second one will filter user requests. Please refer to #101 for https://github.com/hardcoreeng/platform/pull/101/files#diff-a98d95ae7d7f872cd55f123e72db9403e4ec76f5c11ea12b8a8d2f25a0e705be

server/server/src/securityStorage.ts
```diff
@@ -0,0 +1,89 @@
...
+  async findAll<T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>, userId: Ref<Account>): Promise<T[]> {
+    if (this.hierarchy.isDerived(_class, core.class.Space)) return await this.storage.findAll(_class, query)
+    const querySpace = (query as any).space as Ref<Space>
+    const spaces = new Set<Ref<Space>>([...this.publicSpaces, ...this.allowedSpaces.get(userId) ?? []])
+    if (querySpace !== undefined) {
+      if (spaces === undefined || spaces.size === 0) throw new Error('Access denied')
+      if (!spaces.has(querySpace)) throw new Error('Access denied')
+    } else {
+      if (spaces !== undefined) {
+        (query as any).space = [...spaces.values()]
```
haiodo, line 38: I suppose it should be updated after #105

server/server/src/securityStorage.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+    return await this.storage.findAll(_class, query)
+  }
+
+  async tx (tx: Tx, userId: Ref<Account>): Promise<void> {
```
haiodo, line 44: Please use TxProcessor so you will not need to check for transactions, just delegate forward.

server/server/src/securityStorage.ts
```diff
@@ -0,0 +1,89 @@
...
+  private readonly allowedSpaces: Map<Ref<Account>, Set<Ref<Space>>> = new Map<Ref<Account>, Set<Ref<Space>>>()
+  private readonly publicSpaces: Set<Ref<Space>> = new Set<Ref<Space>>()
+
+  constructor (storage: Storage, hierarchy: Hierarchy) {
+    this.storage = storage
+    this.hierarchy = hierarchy
+  }
+
+  async findAll<T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>, userId: Ref<Account>): Promise<T[]> {
+    if (this.hierarchy.isDerived(_class, core.class.Space)) return await this.storage.findAll(_class, query)
```
haiodo, line 30: I think we should just check for Model domain for now and allow any search operations.

haiodo:
@BykhovDenis Please check https://github.com/hardcoreeng/platform/pull/86/files if some bits from this spec will make any seance for initial version of security. Thanks.

server/server/src/securityStorage.ts
BykhovDenis: Yes

server/server/src/security.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  async tx (tx: Tx): Promise<void> {
+    let domain
+    switch (tx._class) {
+      case core.class.TxCreateDoc:
+        domain = this.hierarchy.getDomain((tx as TxCreateDoc<Doc>).objectClass)
+        break
+      case core.class.TxAddCollection:
+        domain = this.hierarchy.getDomain((tx as TxAddCollection<Doc, Emb>).itemClass)
+        break
```
z2sx, line 103: Thinking further `Emb` objects should not belong to a domain, checking domain of embedded object has no sense, we should always check the object which hold collection.

z2sx, line 103: Thinking further `Emb` objects should not belong to a domain, checking domain of embedded object has no sense, we should always check the object which hold collection.

z2sx (changes requested):

server/server/src/security.ts
BykhovDenis: Doesn't ItemClass point to the class of the object containing the collection? Or I can just check `tx.objectSpace == core.space.Model`?

server/server/src/security.ts
```diff
@@ -0,0 +1,109 @@
...
+    const spaces = this.securityModel.getSpaces(this.userId)
+    if (spaces === undefined || spaces.size === 0) throw new Error('Access denied')
+    if (querySpace !== undefined) {
+      if (typeof querySpace === 'string') {
+        if (!spaces.has(querySpace)) throw new Error('Access denied')
+      } else {
+        if ((querySpace.$in?.every((space) => spaces.has(space))) === false) throw new Error('Access denied')
+      }
+    } else {
+      (query as any).space = { $in: [...spaces.values()] }
```
haiodo, line 90: Should we return error to ask space specification here?

server/server/src/security.ts
haiodo: We do not have objectClass for TxAddCollection for now, so I've do same to check domain from itemClass, and it should be same as object class which hold collection.

server/server/src/security.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    return await this.storage.findAll(_class, query)
+  }
+
+  async tx (tx: Tx): Promise<void> {
+    let domain
+    switch (tx._class) {
+      case core.class.TxCreateDoc:
+        domain = this.hierarchy.getDomain((tx as TxCreateDoc<Doc>).objectClass)
```
haiodo, line 99: Please also check for TxUpdateDoc, TxUpdateCollection

server/server/src/security.ts
z2sx: if `itemClass` is "object class which hold collection" it should be named `objectClass`

haiodo: Nope, itemClass is class of item putted into collection.

server/server/src/security.ts
```diff
@@ -0,0 +1,110 @@
...
+      case core.class.TxUpdateDoc:
+        domain = this.hierarchy.getDomain((tx as TxCreateDoc<Doc>).objectClass)
+        break
+      case core.class.TxAddCollection:
+      case core.class.TxUpdateCollection:
+        domain = this.hierarchy.getDomain((tx as TxAddCollection<Doc, Emb>).itemClass)
+        break
+    }
+    if (domain === DOMAIN_MODEL || domain === DOMAIN_TX) return await this.storage.tx(tx)
+    if (!this.securityModel.checkSecurity(this.userId, tx.objectSpace)) throw new Error('Access denied')
```
z2sx, line 107: Use `PlatformError` everywhere

z2sx (changes requested):

PR #101, opened 2021-06-16
Server rework

haiodo:
Server is reworked.
1. Accept connections with tokens and do token validation.
2. Dummy security entry point.
3. server/tools to create a workspace required by server


packages/core/src/domainFilter.ts
```diff
@@ -0,0 +1,57 @@
+import { Class, Doc, Domain, Ref } from './classes'
+import { Hierarchy } from './hierarchy'
+import { Storage } from './storage'
+import { Tx, txObjectClass } from './tx'
+
+/**
+ * Will do not pass transactions with specified domain.
```
haiodo, line 7: Need fix comment.

PR #105, opened 2021-06-16
Updated `documentquery` with operators like `$in`

BykhovDenis:


packages/core/src/storage.ts
```diff
@@ -13,10 +13,23 @@
...
 //
 
-import type { Class, Doc, Ref } from './classes'
+import type { Class, Doc, Obj, Ref } from './classes'
 import type { Tx } from './tx'
 
-export type DocumentQuery<T extends Doc> = Partial<T>
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type QuerySelector<T> = {
+  $gt?: T
```
haiodo, line 21: Why we need any other then $in?

BykhovDenis: I don't know) Some overengineering, because I think `operators like $in` is more then only `$in`)

BykhovDenis: I can delete it, and add when we necessary

packages/core/src/memdb.ts
```diff
@@ -69,9 +69,16 @@ class MemDb {
     if (Object.prototype.hasOwnProperty.call(query, '_id')) {
       if (query._id === undefined) {
         result = []
-      } else {
+      } else if (typeof query._id === 'string') {
         const obj = this.objectById.get(query._id)
         result = obj !== undefined ? [obj] : []
+      } else {
```
haiodo, line 75: This should be as separate PR.

PR #106, opened 2021-06-17
Implemented `$in` query for `memdb`

BykhovDenis:


packages/core/src/__tests__/memdb.test.ts
```diff
@@ -65,6 +69,10 @@ describe('memdb', () => {
       kind: 1
     })
     expect(result.length).toBe(0)
+    const multipleParam = await model.findAll(core.class.Doc, {
```
haiodo, line 72: @BykhovDenis  @z2sx We have some troubles here in case we do search for core.class.Doc, 
assume even for Tx/Memdb stores we should search for all stores and do combine. And right now I think it will not work since it will choose a domain and do search.

Assume we have more domains, some in mongo, some in elastic it will trigger a full search.
Should we allow this?

packages/core/src/storage.ts
```diff
@@ -13,10 +13,23 @@
...
-import type { Class, Doc, Ref } from './classes'
+import type { Class, Doc, Obj, Ref } from './classes'
 import type { Tx } from './tx'
 
-export type DocumentQuery<T extends Doc> = Partial<T>
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type QuerySelector<T> = {
+  $gt?: T
+  $gte?: T
+  $in?: T[]
```
haiodo, line 23: I think we should remove or comment all except $in

packages/core/src/storage.ts
```diff
@@ -13,10 +13,23 @@
...
+  $lt?: T
+  $lte?: T
+  $nin?: T[]
+}
+export type ObjQueryType<T> = T extends Obj ? DocumentQuery<T> : T | QuerySelector<T>
+
+export type DocumentQuery<T> = {
```
haiodo, line 30: I think we also should omit Collection types from here.

packages/core/src/__tests__/memdb.test.ts
z2sx: Could work for test

haiodo (approved):

PR #112, opened 2021-06-20
Kanban layout

SasLord:


plugins/workbench-impl/src/components/icons/Calendar.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+  export let size: number = 24
+  const fill: string = 'var(--theme-caption-color)'
+</script>
+
+<svg width={size} height={size} viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+  <g {fill}>
```
z2sx, line 7: Please move `fill` to `svg` tag everywhere.

z2sx (changes requested):

PR #148, opened 2021-06-29
Like queries

BykhovDenis:
Added `$like` query for wildcard search

Signed-off-by: Denis Bykhov bykhov.denis@gmail.com


haiodo:
@BykhovDenis Please add descriptions to your PRs. Also Please Add Closes #issueId

packages/core/src/__tests__/memdb.test.ts
```diff
@@ -55,7 +55,7 @@ describe('memdb', () => {
     })
     expect(first.length).toBe(1)
     const second = await model.findAll(core.class.Class, {
-      _id: { $in: [txes[1].objectId as Ref<Class<Obj>>, txes[3].objectId as Ref<Class<Obj>>] }
+      _id: { $in: [txes[1].objectId as Ref<Class<Obj>>, txes[3].objectId as Ref<Class<Obj>>], type: '$in' }
```
haiodo, line 58: Why we need type: '$in'?

packages/core/src/__tests__/memdb.test.ts
```diff
@@ -68,11 +68,43 @@ describe('memdb', () => {
...
+    for (const tx of txes) hierarchy.tx(tx)
+    const model = new ModelDb(hierarchy)
+    for (const tx of txes) await model.tx(tx)
+    const expectedLength = txes.filter(tx => tx.objectSpace === core.space.Model).length
+    const without = await model.findAll(core.class.Doc, {
+      space: { $like: core.space.Model, type: '$like' }
+    })
+    expect(without).toHaveLength(expectedLength)
+    const begin = await model.findAll(core.class.Doc, {
+      space: { $like: '*Model', type: '$like' }
```
haiodo, line 87: Why we need type: $like?

packages/core/src/memdb.ts
```diff
@@ -68,26 +88,13 @@ class MemDb extends TxProcessor {
...
-        }
-      }
+      result = this.getByIdQuery(query)
     } else {
       result = this.getObjectsByClass(_class)
     }
 
     for (const key in query) {
-      if (key === '_id') continue
+      if (key === '_id' && (query._id as any)?.$like === undefined) continue
```
haiodo, line 97: Why not allow like for _id?

Model had classes, so I could search for _id: {$like: 'class:core:%'} and will retrieve all classes from plugin?

packages/core/src/predicate.ts
```diff
@@ -31,12 +31,35 @@ const predicates: Record<string, PredicateFactory> = {
       }
       return result
     }
+  },
+
+  $like: (query: string, propertyKey: string): Predicate => {
+    const likeSymbol = '*'
```
haiodo, line 37: Should it be '%'?

packages/core/src/predicate.ts
```diff
@@ -31,12 +31,35 @@ const predicates: Record<string, PredicateFactory> = {
...
+      }
+      return result
+    }
   }
 }
 
 export function isPredicate (o: Record<string, any>): boolean {
   const keys = Object.keys(o)
-  return keys.length === 1 && keys[0].startsWith('$')
+  return keys.length === 2 && ((keys[0].startsWith('$') && keys[1] === 'type') || (keys[1].startsWith('$') && keys[0] === 'type'))
```
haiodo, line 62: Could we make it work without type? I'm not sure why we need it.

haiodo (changes requested):
I'm agains adding `type:'$in'`. Please think how to make it without `type`

packages/core/src/memdb.ts
BykhovDenis: `like` allowed for `_id`, recheck condition please, we get objects by class and then check `_id` as any other `query` field

packages/core/src/predicate.ts
BykhovDenis: I can replace it, but I think '*' is be better here

packages/core/src/predicate.ts
BykhovDenis: I need this to check the type and not convert to `any`. I have not been able to find a better way to get a type from a union inherited from a one base type. I can use functions that check the type by key, and then cast  type by `as`, but I'm not sure if this would be more correct.

haiodo:
```suggestion
  return keys.length >= 1 && keys[0].startsWith(‘$’)
```

haiodo: Why we need any changes here?

packages/core/src/__tests__/memdb.test.ts
```diff
@@ -73,6 +73,38 @@ describe('memdb', () => {
...
+    const zero = await model.findAll(core.class.Doc, {
+      space: { $like: 'Model' }
+    })
+    expect(zero).toHaveLength(0)
+    const end = await model.findAll(core.class.Doc, {
+      space: { $like: 'space:core.M*' }
+    })
+    expect(end).toHaveLength(expectedLength)
+    const mid = await model.findAll(core.class.Doc, {
+      space: { $like: '*M*de*' }
```
haiodo, line 99: $like should allow only prefix and postfix?

packages/core/src/memdb.ts
```diff
@@ -56,6 +56,21 @@ class MemDb extends TxProcessor {
     return result
   }
 
+  private getByIdQuery (query: DocumentQuery<Doc>): Doc[] {
+    const result = []
+    if (query._id === 'string') {
+      const obj = this.objectById.get(query._id)
+      if (obj !== undefined) result.push(obj)
+    } else if (query._id !== undefined && '$in' in query._id) {
+      const ids = query._id.$in
```
haiodo, line 65: Should it also support $like here?

packages/core/src/storage.ts
```diff
@@ -17,16 +17,20 @@ import type { Class, Doc, Ref } from './classes'
 import type { Tx } from './tx'
 
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
-export type QuerySelector<T> = {
-  $in?: T[]
```
haiodo, line 21: Just make it optional. Why we need different ones?
Screenshot 2021-06-30 at 17 50 47

packages/core/src/__tests__/memdb.test.ts
BykhovDenis: Allowed any position, why should we restrict the use? If we used only a prefix or only a postfix, then in the future it would be possible to optimize the algorithms (in particular for `MemDB`), but for a query like `*any*` it will not make sense

packages/core/src/memdb.ts
BykhovDenis: No, here get object by like id impossible, if query includes _id $like, we get objects by class, then filtering

packages/core/src/storage.ts
BykhovDenis: For restrict multiple queryselector

packages/core/src/__tests__/memdb.test.ts
haiodo: ok, i've found sql support putting it inside. It also allow to match for multiple characters and to just one character.

https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like

packages/core/src/memdb.ts
haiodo: Nope, based on code if `Object.prototype.hasOwnProperty.call(query, '_id')` it will go with this branch and will return empty every time.

packages/core/src/predicate.ts
haiodo: Let's @z2sx to decide.

haiodo: Could you please reduce complexity of new code, please use CodeMetrics to estimations.

server/mongo/src/query.ts
```diff
@@ -35,6 +35,12 @@ export function toMongoQuery<T extends Doc> (
   query: DocumentQuery<T>
 ): FilterQuery<T> {
   const mongoQuery: FilterQuery<Doc> = query as FilterQuery<Doc>
+  for (const key in query) {
+    const value = query[key]
+    if (typeof value === 'string') continue
```
haiodo, line 40: value could be number, better to check if `typeof value === 'object'

haiodo:
@BykhovDenis @z2sx I think we should also support escaping of our special symbols. 

 \* or \% to allow passing * or %. 

Thoughts?

packages/core/src/memdb.ts
haiodo: Why not updated?

server/mongo/src/query.ts
```diff
@@ -44,7 +50,19 @@ export function toMongoQuery<T extends Doc> (
...
   }
+
+  console.log(mongoQuery)
   return mongoQuery
 }
+
+function translateQuery<P extends keyof T, T extends Doc> (value: QuerySelector<P>): any {
+  const result: any = {
+  }
+  if (value.$in !== undefined) result.$in = value.$in
```
haiodo, line 63: I think you just need to copy value selector and update $like if present. In this case in future it potentially will be not required to update this piece of code.

server/server/src/security.ts
```diff
@@ -99,12 +99,17 @@ export class SecurityClientStorage implements Storage {
...
         if (!spaces.has(querySpace)) throw new PlatformError(new Status(Severity.ERROR, Code.AccessDenied, {}))
       } else {
-        if (querySpace.$in?.every((space) => spaces.has(space)) === false) {
+        if ((querySpace.$in?.every((space) => spaces.has(space))) === false) {
           throw new PlatformError(new Status(Severity.ERROR, Code.AccessDenied, {}))
         }
+        if (querySpace.$like !== undefined) {
+          const like = querySpace.$like
+          const querySpaces = querySpace.$in ?? [...spaces.values()]
+          query.space = { $in: querySpaces.filter(p => p.match(like)) }
```
haiodo, line 108: This will not work, since like is with our format, and match accept regular expression.

haiodo (changes requested):
Few more comments

packages/core/src/predicate.ts
```diff
@@ -31,17 +31,46 @@ const predicates: Record<string, PredicateFactory> = {
...
+  }
+}
+
+function checkValue (query: string, value: string): boolean {
+  const likeSymbol = '*'
+  const searchValues = query.split(likeSymbol)
+  let startIndex = 0
+  for (const searchValue of searchValues) {
+    const index = value.indexOf(searchValue, startIndex)
+    if (index === -1 || (searchValues[0] !== '' && index > 0 && startIndex === 0)) {
```
haiodo, line 54: Could you please comment why we need second if case?

packages/core/src/storage.ts
```diff
@@ -19,16 +19,22 @@ import type { Tx } from './tx'
...
 }
 
-export type DocumentQuery<T extends Doc> = Query<T>
-
 export interface Storage {
   findAll: <T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>) => Promise<T[]>
   tx: (tx: Tx) => Promise<void>
 }
+
+export function likeToRegExp (value: string): RegExp {
```
haiodo, line 36: @BykhovDenis Let's not leak abstraction, we need a function to check if value is matches like.
So le't do it, match only. We do not need to return regexp, since If you would like to change implementation to custom one like implementation, you will also need to rewrite all places where it is used.

server/mongo/src/query.ts
```diff
@@ -44,7 +50,18 @@ export function toMongoQuery<T extends Doc> (
 
   // Find by all classes.
   if (classes.length > 1) {
-    mongoQuery._class = { $in: classes.map((cl) => cl as Ref<Class<Doc>>) }
+    mongoQuery._class = { $in: classes.map((cl) => cl) }
   }
+
+  console.log(mongoQuery)
```
haiodo, line 56: Please remove logging.

PR #149, opened 2021-06-30
Derived data

haiodo:
Derived Data processor implementation.

If document is described this way
```typescript
interface Task {
 shortId: string
 title: string
 description: string
}

// Following DerivedDataDescriptor will cause automatic creation of Task title index.

builder.createDoc(core.class.DerivedDataDescriptor, {
  sourceClass: testIds.class.Task,
  targetClass: core.class.Title,
  initiValue: {},
  rules: [
    { sourceField: 'shortId', targetField: 'title' }
  ]
})
```

Processor automatically track descriptors and perform update of required derived data.


BykhovDenis (approved):

sumboid (approved):

packages/core/src/derived/processor.ts
```diff
@@ -0,0 +1,237 @@
...
+    return results
+  }
+
+  private async refreshDerivedData (d: Descr, modifiedBy: Ref<Account>): Promise<void> {
+    // Perform a full rebuild of derived data of required type.
+    const ops = withOperations(modifiedBy, this.storage)
+
+    // we need to find all objects affected
+
+    const allDD = await this.storage.findAll(d.sourceClass, {})
```
haiodo, line 215: As discussed on today call, we need API for partial data processing, since it could be a lot of data. #165 could be place to discuss it.

packages/core/src/derived/processor.ts
```diff
@@ -0,0 +1,237 @@
...
+      await this.updateDescriptor(tx.objectId as Ref<Descr>, tx.modifiedBy)
+      return
+    }
+
+    const descriptors = this.descrs.getByClass(tx.objectClass)
+    const ops = withOperations(tx.modifiedBy, this.storage)
+
+    // We expect storage is alrady updated before derived data is being processed.
+    const doc: CachedDoc = {
+      resolve: async () => (await this.storage.findAll(tx.objectClass, { _id: tx.objectId }))[0]
```
haiodo, line 95: Right now I've decided to take source object in case of it update for processing.
Potentially we could check if rules has effect on change and only in this case take object from underline storage.

PR #150, opened 2021-06-30
Fix `Tooltip` component, `Root` resize

SasLord:


packages/ui/src/components/Tooltip.svelte
```diff
@@ -23,7 +23,13 @@
 
 <div class="container">
   <div class="trigger"><slot/></div>
-  <div class="tooltip {direction}"><Label label={label}/></div>
+  <div class="tooltip {direction}">
+    <Label label={label}/>
```
haiodo, line 27: Why two <Label> components?

packages/ui/src/components/Tooltip.svelte
```diff
@@ -64,19 +70,21 @@
...
         height: 15px;
-        background: linear-gradient(-45deg, rgba(44, 44, 52, 1) 50%, rgba(44, 44, 52, 0) 40%);
+        background: linear-gradient(-45deg, var(--theme-tooltip-color) 50%, var(--theme-tooltip-color) 40%);
+        border: 1px solid var(--theme-bg-accent-color);
         border-radius: 0 0 3px;
       }
 
       &.top::after, &.bottom::after {
         left: 50%;
-        margin-left: -7.5px;
```
haiodo, line 79: Ouch, I'm not html guru, 
but use of 0.5px is a good idea?

haiodo (approved):

PR #162, opened 2021-07-01
`sort`&`limit` queries

### LINKED ISSUE #161: Support Limit for `findAll`.
We should have support for limit capability.

BykhovDenis:
Added `sort` and `limit` options for all DB
Close #161


packages/query/src/__tests__/query.test.ts
```diff
@@ -13,10 +13,11 @@
 // limitations under the License.
 //
 
-import type { Class, Client, Doc, DocumentQuery, Obj, Ref, Space, Tx, TxCreateDoc } from '@anticrm/core'
+import type { Class, Client, Doc, DocumentQuery, FindOptions, Obj, Ref, Space, Tx, TxCreateDoc } from '@anticrm/core'
 import core, { createClient, DOMAIN_TX, Hierarchy, ModelDb, TxDb, withOperations } from '@anticrm/core'
 import { genMinModel as getModel } from '@anticrm/core/src/__tests__/minmodel'
 import { LiveQuery } from '..'
+import { SortingOrder } from '../../../core/src'
```
haiodo, line 20: Please use node module dep

packages/core/src/memdb.ts
```diff
@@ -36,6 +36,21 @@ function findProperty (objects: Doc[], propertyKey: string, value: any): Doc[] {
   return result
 }
 
+export function resultSort<T extends Doc> (result: T[], sortOptions: SortingQuery<T>): void {
+  const sortFunc = (a: any, b: any): number => {
+    for (const key in sortOptions) {
```
haiodo, line 41: Probable we need some Comparator registry based on attribute type here.

packages/core/src/storage.ts
```diff
@@ -28,7 +28,22 @@ export type Query<T> = {
...
+}
+
+export enum SortingOrder {
+  Ascending = 1,
+  Descending = -1
+}
+
 export interface Storage {
-  findAll: <T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>) => Promise<T[]>
+  findAll: <T extends Doc>(_class: Ref<Class<T>>, query: DocumentQuery<T>, options?: FindOptions<T>) => Promise<T[]>
```
haiodo, line 47: I've have only one obstacle here, we need to know a real size of response, before limiting and we need to pass it somehow.

In v0.3 I've did this by passing an handler in FindOptions, probable we need to change T[] to some Result<T> extending array with some extra options, like real size, sorting enabled, etc. 

Thoughts?

BykhovDenis: изображение
What about this?

haiodo: It will break all existing code, could you check how regexp match return array with extra props.

packages/query/src/index.ts
```diff
@@ -91,4 +119,58 @@ export class LiveQuery extends TxProcessor implements Client {
...
+
+    if (q.options?.limit !== undefined && q.result.length > q.options.limit) {
+      if (q.result[q.options?.limit]._id === updatedDoc._id) {
+        const res = await this.findAll(q._class, q.query, q.options)
+        q.result = res
+        q.total = res.total
+        q.callback(res)
+        return
+      }
+      if (q.result.pop()?._id !== updatedDoc._id) q.callback(Object.assign(q.result, { total: q.total }))
```
haiodo, line 171: You you please comment on this case?

BykhovDenis: Checking whether the updated document is within the limit

packages/query/src/index.ts
```diff
@@ -104,4 +132,58 @@ export class LiveQuery extends TxProcessor implements Client {
...
+      }
+    }
+    updatedDoc.modifiedBy = tx.modifiedBy
+    updatedDoc.modifiedOn = tx.modifiedOn
+  }
+
+  private sort (q: Query, tx: TxUpdateDoc<Doc>): void {
+    const sort = q.options?.sort
+    if (sort === undefined) return
+    let needSort = sort.modifiedBy !== undefined || sort.modifiedOn !== undefined
```
sumboid, line 153: This condition is a bit unclear

packages/query/src/__tests__/query.test.ts
```diff
@@ -176,6 +176,42 @@ describe('query', () => {
...
 
+  it('limit and sorting', async (done) => {
+    const storage = await getClient()
+
+    const limit = 1
+    let attempt = 0
+    let doneCount = 0
+
+    const query = withOperations(core.account.System, new LiveQuery(storage))
+    query.query<Space>(core.class.Space, { private: true }, (result) => {
```
haiodo, line 187: Could `new Promise((resolve)=> ...) `make it a bit more clear, so we could not use doneCount in this case.

packages/core/src/query.ts
```diff
@@ -25,3 +26,14 @@ export function findProperty (objects: Doc[], propertyKey: string, value: any):
   }
   return result
 }
+
+export function resultSort<T extends Doc> (result: T[], sortOptions: SortingQuery<T>): void {
+  const sortFunc = (a: any, b: any): number => {
+    for (const key in sortOptions) {
+      const result = typeof a[key] === 'string' ? a[key].localeCompare(b[key]) : a - b
```
haiodo, line 33:
```suggestion
      const result = typeof a[key] === 'string' ? a[key].localeCompare(b[key]) : a[key] - b[key]
```

haiodo (changes requested):
Some fixes required

haiodo (approved):

PR #171, opened 2021-07-02
Security use model

### LINKED ISSUE #158: Security should use Model, #157

BykhovDenis:
close #158

### DIFF
--- a/server/server/src/security.ts
+++ b/server/server/src/security.ts
@@ -19,24 +19,24 @@ import { component, Component, PlatformError, Severity, Status, StatusCode } fro
 
 export class SecurityModel extends TxProcessor {
   private readonly hierarchy: Hierarchy
-  private readonly model: ModelDb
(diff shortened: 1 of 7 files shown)


server/server/src/security.ts
```diff
@@ -18,27 +19,39 @@ import { component, Component, PlatformError, Severity, Status, StatusCode } fro
...
   private readonly publicSpaces: Set<Ref<Space>> = new Set<Ref<Space>>()
 
-  constructor (hierarchy: Hierarchy) {
+  constructor (hierarchy: Hierarchy, model: ModelDb) {
     super()
     this.hierarchy = hierarchy
+    this.model = model
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    model.findAll(core.class.Space, {}).then(spaces => {
```
haiodo, line 32: Not a good style, could you please create a static async create function and do all actions inside it. So it will be possible to properly test and be same all is initialised before it will be used.

haiodo (approved):

PR #178, opened 2021-07-07
Initial FSM implementation

### LINKED ISSUE #154: FSM plugin
```ts
interface FSM extends Doc {
  name: string
  application: Ref<Application>
  isTemplate: boolean
}

interface Transition extends Doc {
  from: Ref<State>
  to: Ref<State>
  fsm: Ref<FSM>
}

interface WithFSM extends Doc {
  fsm: Ref<FSM>
}

interface FSMItem extends Doc {
  fsm: Ref<WithFSM>
  state: Ref<State>
  item: Ref<VDoc>
  clazz: Ref<Class<VDoc>>
}

interface State extends Doc {
  name: string
  fsm: Ref<FSM>
}
```

Diagram
To build project kanban we need to determine linked FSM and gather States, FSMItems. It would be enough to render board. To limit user actions we should also check transitions.

Additional cases we probably should take into account:
1. Getting state of `Task`. Not sure it would work fine, without schema changes we need to find `FSMItems` refers to `Task`, after that we can even find related FSM.
2. ...

sumboid:
Probably be slightly changed after actual integration
Closes #154


plugins/fsm/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import { plugin } from '@anticrm/platform'
+import type { Plugin, Service } from '@anticrm/platform'
+import type { Doc, Ref, Class } from '@anticrm/core'
+import type { Application } from '@anticrm/workbench'
+
+export interface FSM extends Doc {
```
haiodo, line 21: What about rename FSM to StateModel?

plugins/fsm-impl/src/components/App.svelte
```diff
@@ -0,0 +1,187 @@
+<!--
```
haiodo, line 1: How this file is related to FSM?

sumboid: Yeah, good point :)

plugins/fsm/src/index.ts
sumboid: Do you mean just here or overall?

haiodo: Overall

plugins/fsm/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  isTemplate: boolean
+}
+
+export interface Transition extends Doc {
+  from: Ref<State>
+  to: Ref<State>
+  fsm: Ref<FSM>
+}
+
+export interface WithFSM extends Doc {
```
haiodo, line 33: Why we need this one?

plugins/fsm/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  fsm: Ref<FSM>
+}
+
+export interface FSMItem extends Doc {
+  fsm: Ref<WithFSM>
```
haiodo, line 38: Why no Ref<FSM>?

PR #182, opened 2021-07-08
Feature/clientimpl update model

### LINKED ISSUE #176: ClientImpl should update model copy in case of model operations are applied
Right now if we construct ClientImpl by `createClient` call and perform creation of model object, it will go on server and will not be available to find from client side.

We need to have test for this case and appropriate fixes.

Please consider having: server/workspace/src/storage.ts:txObjectClass moved to some core place and used to obtain domain for both client and workspace.

BykhovDenis:
close #176


packages/core/src/client.ts
```diff
@@ -71,7 +71,9 @@ export async function createClient (
       txBuffer?.push(tx)
     } else {
       hierarchy.tx(tx)
-      void model.tx(tx)
+      if (tx.objectSpace === core.space.Model) {
+        void model.tx(tx)
```
haiodo, line 75: Please also do same on line 55 ^^^

```typescript
 async tx (tx: Tx): Promise<void> {
    await this.conn.tx(tx)
    // Should also have
    hierarchy.tx(tx)
    if (tx.objectSpace === core.space.Model) {
       void model.tx(tx)
    }
}
```

BykhovDenis: No, because we will have double tx in this case, conn.tx() will call handler....

PR #193, opened 2021-07-09
Tasks shorRefId

BykhovDenis:


packages/core/src/tx.ts
```diff
@@ -117,6 +116,30 @@ export function withOperations<T extends WithTx> (user: Ref<Account>, storage: T
...
+      objectId: generateId(),
+      objectClass: _class,
+      objectSpace: space,
+      attributes
+    }
+    await storage.tx(tx)
+    const result = TxProcessor.createDoc2Doc(tx) as T
+    const objectSpace = (await storage.findAll(core.class.Space, { _id: result.space }, { limit: 1 }))[0]
+    const workspace = objectSpace.name.trim().toUpperCase().replace(/[^a-z0-9]/gim, '_').replace(/[_]/g, '_')
+    await createShortRef(storage, user, result, workspace)
```
haiodo, line 139: For task I've propose to store short Id as part of object itself as well to not perform extra find requests, since it will be good to see it in most places task will be shown.

packages/core/src/shortref.ts
```diff
@@ -30,9 +30,7 @@ export interface ShortRef extends Doc {
 export async function createShortRef<T extends Doc> (
   storage: Storage,
   user: Ref<Account>,
-  space: Ref<Space>,
```
haiodo, line 33: I'm not very like this changes, not sure why.

BykhovDenis: Only about "space" field or all of this?

packages/core/src/tx.ts
```diff
@@ -117,6 +116,30 @@ export function withOperations<T extends WithTx> (user: Ref<Account>, storage: T
     return TxProcessor.createDoc2Doc(tx) as T
   }
 
+  result.createDocWithShortRef = async <T extends Doc>(
```
haiodo, line 119: Let's have separate createShortRef if you would like to have it here.

packages/core/src/shortref.ts
haiodo: If you like just to simplify passing 3 params instead of 1 param, let's create an interface with just 3 this params, in this case you will be able to pass entire doc, but still will be possible to call this method without passing a doc, but just required parameters.

haiodo (approved):

PR #200, opened 2021-07-12
Task kanban

### LINKED ISSUE #197: Move Task interface to task plugin instead of task-impl
Move Task interface to task plugin instead of task-impl

### LINKED ISSUE #198: Task Status is shown as number instead of Enum value.
Please fix TaskStatus presenter to show a proper value.
Task Status is shown as number instead of Enum value.

BykhovDenis:
close #197 , close #198

### DIFF
--- a/plugins/task/src/index.ts
+++ b/plugins/task/src/index.ts
@@ -16,7 +16,7 @@
 import { plugin } from '@anticrm/platform'
 import type { Asset, IntlString } from '@anticrm/status'
 import type { Plugin, Service } from '@anticrm/platform'
-import type { Account, Doc, Ref, Space } from '@anticrm/core'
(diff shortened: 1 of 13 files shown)


plugins/task/src/index.ts
```diff
@@ -14,14 +14,29 @@
...
+import type { Asset, IntlString } from '@anticrm/status'
 import type { Plugin, Service } from '@anticrm/platform'
-import type { Space } from '@anticrm/core'
+import type { Account, Doc, Ref, Space } from '@anticrm/core'
 
 export interface Project extends Space {}
 
 export interface TaskService extends Service {}
 
+export const TaskStatuses = {
```
haiodo, line 25: Why not use regular enum with string values?

BykhovDenis: I think extend it additional fields (color as example)

PR #209, opened 2021-07-13
Make server working again

### LINKED ISSUE #204: Make Server working
Need a server to retrieve accountId() decoded from token for a proper usage.

haiodo:
1. Make server working again
2. TxOperations now has accountId()
3. Platform Client has async accountId() to retrieve current authenticated accountId from server.
3. Plugin Client has accountId() to return current accountId.
4. Some re-formattings

Closes #204


dev/prod/webpack.config.js
```diff
@@ -151,7 +152,10 @@ module.exports = {
     new MiniCssExtractPlugin({
       filename: '[name].css'
     }),
-    new Dotenv()
+    new Dotenv(),
+    new DefinePlugin({
+      'process.env.CLIENT': JSON.stringify(process.env.CLIENT)
```
haiodo, line 157: In case Dotenv doesn't have variable we could pass it from NODE, this is used for `rushx dev-server` to start web ui and connect to server.

models/core/src/derived.ts
```diff
@@ -71,7 +71,7 @@ export class TBackReference extends TDerivedData implements BackReference {
   sourceObjectClass!: Ref<Class<Doc>>
 }
 
-@Model(core.class.ShortRef, core.class.Title, DOMAIN_REFERENCE)
+@Model(core.class.ShortRef, core.class.Title, DOMAIN_TITLE)
```
haiodo, line 74: Short ref is title, so should be in titles

BykhovDenis (approved):

PR #213, opened 2021-07-14
Task cards

BykhovDenis:


plugins/task/src/index.ts
```diff
@@ -31,11 +31,18 @@ export const TaskStatuses = {
...
+  shortRefId: Ref<ShortRef>,
+  name: string,
+  description: string,
+  assignee?: Ref<Account>,
+  status: TaskStatus,
+  subtasks: Array<Subtask>,
+  commentSpace: Ref<Space>
+}
+
+export interface Subtask {
```
haiodo, line 43: I'm not like Subtask name, since it is not a subtask.

Let's make it `CheckListItem` and put them inside object itself. I've not think we need them outside of Task.

PR #220, opened 2021-07-16
Extract table to ui plugin

sumboid:


BykhovDenis (approved):

packages/ui/src/components/Table.svelte
```diff
@@ -0,0 +1,123 @@
...
+
+  import { AnySvelteComponent } from '../types'
+  import Label from './Label.svelte'
+
+  interface Cell {
+    component: AnySvelteComponent
+    props?: Object
+  }
+
+  interface Column {
```
haiodo, line 26: Should we put this files to types.ts and reuse for type safety?

plugins/task-impl/src/components/TableView.svelte
```diff
@@ -14,128 +14,41 @@
...
-    unsubscribe()
-    unsubscribe = client.query(_class, { space: currentSpace }, (result) => data = result)
-  }
+  let prevSpace: Ref<Space> | undefined
 
-  onDestroy(() => {
-    unsubscribe()
-  })
+  const _class: Ref<Class<Doc>> = task.class.Task
+  const columns = [
```
haiodo, line 31: Should we use Column from ui to be sure all is type safe?

packages/ui/src/components/Table.svelte
sumboid: @haiodo Probably I misunderstood you, we will get an error if try to provide mismatching columns definition, isn't that enough?
What I really would like to achieve is having parametric component to make sure we don't use non-existing doc fields, like:
```
export let data: T[]

interface ColumnProp<T> {
  property: string
  key?: keyof T
  value?: any
}
```

PR #222, opened 2021-07-16
Task edit

BykhovDenis:


packages/ui/src/components/Progress.svelte
```diff
@@ -18,13 +18,13 @@
   export let max: number = 100
   export let color: string = '#50BCF9'
 
-  const proc: number = (max - min) / 100
```
haiodo, line 21: Better to have let proc, since liner will show error here.

BykhovDenis: We need calculate `proc` dynamically because `max` can be changed external without component redraw. As example when we change checkTasks quantity

haiodo (approved):

PR #223, opened 2021-07-16
Extract kanban

sumboid:


packages/ui/src/components/kanban/Panel.svelte
```diff
@@ -12,28 +12,35 @@
...
   let collapsed: boolean = false
 </script>
 
-<section class="panel" on:dragover on:drop class:collapsed={collapsed}>
-  <div class="header" style="background-color: {color}" on:click={() => collapsed = !collapsed}>
-    {#if collapsed !== true}<div class="title"><Label label={title}/></div>{/if}
+<section class="panel" on:dragover on:drop class:collapsed>
+  <div
+    class="header"
+    style="background-color: {color}"
```
BykhovDenis, line 30: Can you please fix style as figma design, #222 as example

sumboid: @BykhovDenis  I'm not sure it is actual mockup, at least I see multiple versions, interactive one is different: https://www.figma.com/proto/coVb3AHZaVKGERM806zHGv/Platform-Prototype?page-id=0%3A1&node-id=74%3A68315&viewport=-30626%2C-4086%2C0.8550704121589661&scaling=contain

@z2sx @MarinaMatijaca which one should be implemented?
2021-07-17-194122_371x730_scrot
or
2021-07-17-194111_282x641_scrot

PR #232, opened 2021-07-20
Security space

### LINKED ISSUE #226: Space Security
We need to allow space operations to persons belong to spaces.

1. If space is public one, we should allow to join for anyone.
2. If space user is in members, we should allow to add more persons or leave.
3. We should send space objects to persons only if they are belong to space.

BykhovDenis:
Simple security impl for space txes
close #226


server/server/src/security.ts
```diff
@@ -62,28 +63,71 @@ export class SecurityModel extends TxProcessor {
...
-          accountSpaces.add(tx.objectId as Ref<Space>)
-        }
+      const spaceTx = tx as TxUpdateDoc<Space>
+      this.changeSpacePrivate(spaceTx)
+      this.changeSpaceMembers(spaceTx)
+    }
+  }
+
+  changeSpaceMembers (spaceTx: TxUpdateDoc<Space>): void {
+    const member = spaceTx.operations?.$push?.members
```
haiodo, line 73: Please add $pull support

PR #239, opened 2021-07-21
Chunter Threads support

### LINKED ISSUE #227: Chunter: Thread messages support
We should be capable of staring thread messages.

1. User should be able to start a thread for any message writed.

2. User should be capable of edit of own messages in threads.

haiodo:
1. Chunter Threads support
2. Add DD collection allow to map objects
3. Fix LiveQuery update for dev and server modes
4. Fix DD for dev handled multiple times.

Close #227 

### DIFF
--- a/plugins/chunter-impl/src/components/Comments.svelte
+++ b/plugins/chunter-impl/src/components/Comments.svelte
@@ -13,14 +13,13 @@
 // limitations under the License.
 -->
 <script lang="ts">
-  import { Message as MessageModel } from '@anticrm/chunter'
-  import type { Ref, Space } from '@anticrm/core'
+  import { Message as MessageModel, Comment } from '@anticrm/chunter'
   import { getClient } from '@anticrm/workbench'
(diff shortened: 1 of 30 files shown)


BykhovDenis (approved):

plugins/chunter-impl/src/components/Comments.svelte
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  onDestroy(() => {
+    unsubscribe()
+  })
+</script>
+
+<div class="channel-container" bind:this={div}>
+  {#each comments as m (m._id)}
+    <Message message={m} />
```
sumboid, line 48: `thread={true}` seems to be necessary here

sumboid (changes requested):
If you try to reply to the thread you may notice your message appears at the beginning of thread, is this expected behavior?

plugins/chunter-impl/src/components/Comments.svelte
haiodo: Nice, resolved.

PR #240, opened 2021-07-21
Task assignee account

BykhovDenis:


models/core/src/index.ts
```diff
@@ -43,4 +43,12 @@ export function createModel (builder: Builder): void {
     TReference,
     TShortRef
   )
+
+  builder.createDoc(
```
haiodo, line 47: I don't think we need it in model

packages/core/src/classes.ts
```diff
@@ -100,4 +100,7 @@ export interface Space extends Doc {
   members: Arr<Ref<Account>>
 }
 
-export interface Account extends Doc {}
+export interface Account extends Doc {
+  name: string
```
haiodo, line 104: name is full name? Should it be structured and optional? It will be not so many changes, and will be more useful?

packages/ui/src/components/UserBox.svelte
```diff
@@ -58,43 +53,45 @@
...
+          dispatch('change')
         }}
       />
     {/each}
   </PopupMenu>
   <div class="selectUser">
     <div class="title"><Label label={title} /></div>
     <div class="user">
-      {#if selected}{selected.title}{:else}<Label {label} />{/if}
+      {#if selectedItem}{selectedItem.name}{:else}<Label {label} />{/if}
```
haiodo, line 94: You should also show _id I think.

BykhovDenis: I think that `_id` may be user e-mail or something else, and `name` it be fullName or nickName, or anything that user whant to be showing

haiodo: Yes _id will be email or accountId.

models/core/src/index.ts
BykhovDenis: I think if we have system user account that it need be create

haiodo (approved):

PR #248, opened 2021-07-27
Add component Section

SasLord:


packages/ui/package.json
```diff
@@ -11,6 +11,7 @@
     "lint": "eslint src",
     "lint:fix": "eslint --fix src",
     "format": "prettier --write --plugin-search-dir=. 'src/**/*.{ts*,js*,yml,svelte}' && eslint --fix src",
```
haiodo, line 13: @SasLord  Please just remove \' on format, it works fine on macos. And yaml seems to be not required.

haiodo:
@SasLord Could you check warnings please? 

Screenshot 2021-07-29 at 11 21 37

PR #249, opened 2021-07-27
Derived data fix

BykhovDenis:
Fixed multiple update tx, implemented fullRef with _class

### DIFF
--- a/plugins/plugin-core-dev/src/connection.ts
+++ b/plugins/plugin-core-dev/src/connection.ts
@@ -16,28 +16,39 @@
 import core, {
   Account,
   Class,
+  Client,
+  DerivedDataProcessor,
   Doc,
   DocumentQuery,
   DOMAIN_TX,
   FindResult,
   Hierarchy,
   ModelDb,
+  Obj,
   Ref,
   Storage,
   Tx,
   TxDb,
+  TxProcessor,
   WithAccountId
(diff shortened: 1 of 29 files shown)


plugins/plugin-core-dev/src/connection.ts
```diff
@@ -107,5 +120,19 @@ class ClientImpl implements WithAccountId {
 }
 
 export async function connect (handler: (tx: Tx) => void): Promise<WithAccountId> {
-  return await ClientImpl.create(handler)
+  const client = await ClientImpl.create(handler)
+  const ddProcessor = await DerivedDataProcessor.create(client.model, client.hierarchy, newClientOnlyStorage(client))
```
haiodo, line 124: It is not required, since it is already in core client.

BykhovDenis: Required... Because in core client we don't save dd transaction in DB

haiodo: And we should not save them

haiodo: We just need notification to update live queries

haiodo: Ok, since it server simulation we need it.

haiodo (approved):

PR #269, opened 2021-08-05
Chunter minor fixes

BykhovDenis:


plugins/chunter-impl/src/components/Message.svelte
```diff
@@ -59,23 +57,31 @@
...
+    const today = new Date(new Date().getFullYear(), new Date().getMonth(), new Date().getDate()).getTime()
+    const date = new Date(value).getTime()
+    return date - today > 0
+  }
 </script>
 
 {#await getUser(message.modifiedBy) then user}
   <div class="message-container" on:click={onClick}>
-    <div class="avatar"><img src={user?.avatar ?? ''} alt="Avatar" /></div>
+    <div class="avatar"><img src={user?.avatar ?? ''} alt={user?.name} /></div>
```
haiodo, line 70: Probable we need some default account image in case there is no user one specified.

plugins/chunter-impl/src/components/Replies.svelte
```diff
@@ -13,26 +13,37 @@
...
+  export let replies: Ref<Account>[] = []
 
   const shown: number = 4
-  const showReplies: Array<string> = []
-  for (let i = 0; i < shown; i++) {
-    showReplies.push(replies[i])
+  let showReplies: Array<Account> = []
+
+  async function getUsers (): Promise<Array<Account>> {
+    const users = await client.findAll(core.class.Account, { _id: { $in: replies } })
```
haiodo, line 30: Should we have limit passed to findAll instead?

haiodo (approved):

PR #272, opened 2021-08-10
Notification

BykhovDenis:


packages/core/src/predicate.ts
```diff
@@ -43,6 +43,28 @@ const predicates: Record<string, PredicateFactory> = {
...
       return result
     }
+  },
+
+  $gt: (query: number, propertyKey: string): Predicate => {
+    return (docs: Doc[]): Doc[] => {
+      const result: Doc[] = []
+      for (const doc of docs) {
+        const value = (doc as any)[propertyKey]
+        if (value - query > 0) result.push(doc)
```
haiodo, line 53: Should it be just for number formats?

packages/query/src/index.ts
```diff
@@ -134,6 +135,10 @@ export class LiveQuery extends TxProcessor implements Storage, Queriable {
       const updatedDoc = q.result.find((p) => p._id === tx.objectId)
       if (updatedDoc !== undefined) {
         await this.doUpdateDoc(updatedDoc, tx)
+        if (!this.match(q, updatedDoc)) {
+          q.result.splice(0, q.result.indexOf(updatedDoc))
+          q.total--
```
haiodo, line 140: good catch

plugins/chunter-impl/src/components/ChannelView.svelte
```diff
@@ -37,8 +39,10 @@
   let query: QueryUpdater<Message> | undefined
 
   $: if (currentSpace !== undefined) {
-    query = client.query(query, chunter.class.Message, { space: currentSpace }, (result) => {
+    query = client.query(query, chunter.class.Message, { space: currentSpace }, async (result) => {
       messages = result
+      const notificationP = await getPlugin(notificationPlugin.id)
```
haiodo, line 44: Please move to constants

plugins/notification-impl/src/index.ts
```diff
@@ -0,0 +1,331 @@
...
+  const accountId = client.accountId()
+
+  const markAsRead = async <T extends Doc>(objectClass: Ref<Class<T>>, space: Ref<Space>): Promise<void> => {
+    const query = {
+      objectClass: objectClass,
+      space: space,
+      client: accountId
+    }
+    // eslint-disable-next-line no-void
+    void client.findAll(notification.class.LastViewed, query).then((subscribes) => {
```
haiodo, line 36: why not use await for client.findAll?

packages/core/src/predicate.ts
BykhovDenis: Now, yes, I can hardly imagine a situation when you need `gt` not by the number ...

plugins/notification-impl/src/index.ts
BykhovDenis: Faster

plugins/chunter-impl/src/components/ChannelView.svelte
BykhovDenis: Not understood. What exactly should I move?

PR #279, opened 2021-08-14
Separate account

haiodo:
1. Use accountId generated with account host, it should be uniq.
2. Add email to Account 
3. Some small fixes related to it.

### DIFF
--- a/server/server/src/wsserver.ts
+++ b/server/server/src/wsserver.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 //
 
-import core, { Account, newTxCreateDoc, Ref, Storage, Tx } from '@anticrm/core'
+import core, { Account, newTxCreateDoc, Ref, Tx, WithAccountId } from '@anticrm/core'
 import * as gravatar from 'gravatar'
 import { Server, start } from './server'
 import { AccountDetails, decodeToken } from './token'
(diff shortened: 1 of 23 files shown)


server/server/src/wsserver.ts
```diff
@@ -22,18 +22,25 @@ import { assignWorkspace, closeWorkspace, WorkspaceInfo } from './workspaces'
...
     close: async (clientId) => {
       await closeWorkspace(clientId)
     }
   })
   return instance
 }
 function connectClient (
-  serverToken: string
-): (clientId: string, token: string, sendTx: (tx: Tx) => void, close: () => void) => Promise<Storage> {
+  serverToken: string, options?: ServerOptions
```
sumboid, line 42: I'd suggest to set default options:
```ts
options: ServerOptions = {logRequests: false, logTransactions: false}
```

sumboid (approved):

server/server/src/wsserver.ts
haiodo: For me it is same as not defined. Will fix if more options will occur.

BykhovDenis (approved):

server/server/src/wsserver.ts
haiodo: Fixed

PR #283, opened 2021-08-17
Initial actions implementation

### LINKED ISSUE #247: Actions design
# Actions plugin

Action is a server-side flow that can be initialized by Platform user. It is interruptible and can continue execution after server restart. Basically we want next thing: I have a `Make something great` button, by hitting it I invoke chain of asynchronous processing that might require explicit user action to get some info.

### General idea
Let's define action as set of operations (or atoms, not sure which name is better) executing on stack machine. Operation is async js function that takes execution context:
```ts
type Context = {
  pop: () => any
  client: Client // Some client that provides operations that we describe below
}

type Op = async (ctx: Context) => Promise<any>

const Action = new Action()
  .exec(async (ctx: Context) => ctx.client.query('num0')) // stack: [2]
  .exec(async (ctx: Context) => ctx.client.query('num1')) // stack: [3, 2]
  .exec(async (ctx: Context) => ctx.pop() + ctx.pop())    // 3 + 2, stack: [5]
  .build()
  .exec()
```

Additionally we would support goto with conditions and labels. (Of course we can use some `state: any` instead of stack, not sure which is better, maybe we can combine them together.)

So, what profit do we have: after every call we have stack and counter which we can save as execution context and execute action with it:
```ts
const Action = new Action()
  .exec(async (ctx: Context) => ctx.client.query('num0')) // stack: [2]
  .exec(async (ctx: Context) => ctx.client.query('num1')) // stack: [3, 2]
  .exec(async (ctx: Context) => ctx.pop() + ctx.pop())    // 3 + 2, stack: [5]
  .build()
  .exec({counter: 2, stack: [3, 2]}) // call only third op
```

You can take a look at playground: https://replit.com/@sumboid/Action-playground (not finished yet, I use it to test some crazy ideas based on this solution)

### Atomicity of operation transactions
From atomicity perspective, each operation may write to database, but to keep data consistent we need to commit transactions after end of execution. It doesn't require some changes, for beginning we can collect all txes and apply them as soon as executor calls `commit`. Consider next operation example:
1. Write something to DB
2. Wait for some change (literally can take days)
3. Service reboot
4. Need to restart operation and write again

If we write to DB after request we make duplicate what is not correct, so explicit commit is required.

### Client API
As for client api, I think we need next capabilities, most likely it covers all our needs:
1. Get object from storage
2. Create/modify object
3. Wait for object being in specific state, like wait until objects be created, or become modified to match specific condition.

### Action cancelling
It may be required to cancel action, for instance, by user request. For this case we can make client api to be interruptable and also provide Promise wrapper for users if they need to make their long-living promises to be cancellable.
One question there is should we additionally clean up anything? For instance request objects, since they don't make sense after cancel.

### Action composition
It may be useful to call one actions from another. For instance, we can have package with some common actions that can be configured to fit specific case like gather feedback from N users: it can be used almost everywhere and be linked with any object:
```ts
  const gatherFeedbackAction = new Action()
    .exec(async (ctx: Context) => {
      const [id, space, users] = ctx.pop()
      
      createNecessaryRequestObjs(id, space, users)
      
      return [id, space, users]
    })
    .exec(async (ctx: Context) => {
      const [id, space, users] = ctx.pop()

      ctx.waitFor(space, { refTo: id }, (result) => users.every(user => result.some(x => x.ownerId === user)))
    }
    .build() // We need to be able to resolve this action
    
  const syntheticAction = new Action()
    .exec(async (ctx: Context) => {
      const obj = await ctx.client.get('obj_id', 'some_space')
      
      return [obj.id, 'some_space', obj.users.map(x => x.id)]
    })
    .call('actions:package.gatherFeedback')
    .build()
```

## Model
Let's take a look how actions may look from model perspective. First, we need some way to register actions on the server side, probably it could be different service it doesn't matter. As for starting action, we may simply create instance any way we want and service will handle this transaction to start action.

```typescript
// Probably it is redundant
interface Action extends Doc {
  name: string
  description: string
  input: Ref<Class<Doc>>
  resID: Resource
}

interface ActionInstance extends Doc {
  action: Ref<Action>
  target: Ref<Doc>
  input: Ref<Doc>
  state: Enum<'pending' | 'fulfilled' | 'rejected'>
  context: Ref<ActionContext>
  fulfill?: Ref<Doc>
  reject?: Error
}

interface ActionContext extends Doc {
  counter: Array<number>
  stack: Array<any> // Somehow need to define any, because it is literally anything
}

```

sumboid:
Resolves #247


models/chunter/src/index.ts
```diff
@@ -42,6 +43,7 @@ export class TChannel extends TSpace implements Channel {
  */
 @Model(chunter.class.Message, core.class.Doc, DOMAIN_CHUNTER)
 export class TMessage extends TDoc implements Message {
+  isChunterbot!: boolean
```
haiodo, line 46: Why we need isChunterbot, is channel is not enough?

models/chunter/src/index.ts
```diff
@@ -60,6 +63,28 @@ export class TComment extends TDoc implements Comment {
  */
 export function createModel (builder: Builder): void {
   builder.createModel(TChannel, TMessage, TComment)
+  const directMessagesModel: SpacesNavModel<Channel> = {
+    label: chunter.string.DirectMessages,
+    spaceIcon: chunter.icon.Hashtag,
+    spaceClass: chunter.class.Channel,
+    spaceQuery: { direct: true, 'account.starred': { $ne: true } },
```
haiodo, line 70: Why you need account.starred for Chunter bot to be specified?

models/recruiting/src/index.ts
```diff
@@ -108,14 +172,52 @@ export function createModel (builder: Builder): void {
...
-    techInterview: { name: 'Technical interview' },
-    offer: { name: 'Offer' },
-    contract: { name: 'Contract signing' }
+  // Actions
+
+  const factorialId: Ref<Action> = generateId()
+  builder.createDoc(
+    action.class.Action,
+    {
+      name: 'Factorial',
```
haiodo, line 181: Let's remove factorial stuff already.

packages/action/src/action.ts
```diff
@@ -0,0 +1,185 @@
...
+export interface TxFilter {
+  space?: Ref<Space>
+  clazz?: Ref<Class<Doc>>
+  id?: Ref<Doc>
+}
+
+/**
+ * @public
+ */
+export type TxCallback = (tx: TxCreateDoc<Doc> | TxUpdateDoc<Doc>) => Promise<void>
```
haiodo, line 16: You could use ObjectTx

packages/action/src/action.ts
```diff
@@ -0,0 +1,185 @@
+import type { Account, Class, Doc, Ref, Space, Storage, TxCreateDoc, TxUpdateDoc } from '@anticrm/core'
+import { nanoid } from 'nanoid'
```
haiodo, line 2: Why not use generateId() for similar Ids in all places?

plugins/action/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
```
haiodo, line 1: Better to reuse `platform-rig` for defaults

plugins/action/api-extractor.json
```diff
@@ -0,0 +1,17 @@
+/**
+ * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
```
haiodo, line 2: Please reuse `platform-rig`

plugins/calendar-action/src/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    if (tid !== undefined) {
+      clearTimeout(tid)
+    }
+  }
+
+  return res
+}
+
+class DeferredPromise<T> {
```
haiodo, line 49: I think we probable need this some there in core.

plugins/chunter-impl/src/components/Message.svelte
```diff
@@ -54,6 +54,15 @@
   const router = getRouter<WorkbenchRoute>()
 
   async function getUser (userId: Ref<Account>): Promise<Account> {
+    if (message.isChunterbot) {
```
haiodo, line 57: Could we think about some system account id instead?

plugins/calendar-action/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      if (event === undefined) {
+        throw Error(`Event is not found: ${eventID}`)
+      }
+
+      let { endsAt } = event
+      let update = new DeferredPromise<number>()
+
+      const unsub = ctx.subscribe({ id: event._id }, async (tx) => {
+        if (tx._class !== core.class.TxUpdateDoc) {
```
haiodo, line 63: What if event is removed? It will wait forever?

plugins/calendar-action/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    if (tid !== undefined) {
+      clearTimeout(tid)
+    }
+  }
+
+  return res
+}
+
+const waitForEvent = new Action()
```
haiodo, line 49: So all of this is to check for event to be complete for a moment.

I personally think, better to have a boolean field for event named 'isComplete' and derived data to update this field on event updates.

So UI could allow to Move Application from one state to another state in case associated event is marked as isComplete. 

In this case Actions will not be required to be executed on server at all, and we will only need an UI action contribution to create an event.

plugins/calendar-action/src/index.ts
sumboid: Planned to handle this case in different PR. In general this case is complex, it is necessary to allow user to cancel/create new instance of action, and removing or archiving event would lead to action cancel or something like that.

plugins/recruiting-impl/src/components/feedback/FeedbackForm.svelte
```diff
@@ -0,0 +1,77 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// 
+// See the License for the specific language governing permissions and
+// limitations under the License.
+-->
+<script lang="ts">
+  import { Class, Ref } from '@anticrm/core'
+  import { getClient } from '@anticrm/workbench'
+  import { Button, TextArea } from '@anticrm/ui'
+  import { QueryUpdater } from '@anticrm/presentation'
```
haiodo, line 19: import type

plugins/recruiting-impl/src/components/vacancies/ApplicantCard.svelte
```diff
@@ -16,41 +16,150 @@
...
 
   export let doc: Applicant
-  let candidate: Candidate | undefined
 
   const client = getClient()
-  let lq: QueryUpdater<Candidate> | undefined
 
-  $: {
-    lq = client.query(lq, recruiting.class.Candidate, { _id: doc.item as Ref<Candidate> }, ([first]) => {
+  console.log(doc)
```
haiodo, line 35: logging

plugins/recruiting-impl/src/components/vacancies/ApplicantCard.svelte
```diff
@@ -16,41 +16,150 @@
...
+    ([first]) => {
       candidate = first
-    })
+    }
+  )
+
+  let state: State | undefined
+  let stateQ: QueryUpdater<State> | undefined
+  $: stateQ = client.query(stateQ, fsm.class.State, { _id: doc.state }, (res) => {
+    console.log(res)
```
haiodo, line 50: - logging

server/server/config/jest.config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "@anticrm/platform-rig/profiles/default/config/jest.config.json",
+  "coverageThreshold": {
```
haiodo, line 3: TODO: Create Ticket to fix coverage.

sumboid: Done: #374

server/server/src/actions/runtime.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  async tx (_clientId: string, tx: Tx): Promise<void> {
+    return await this.txHandlers[tx._class]?.(tx)
+  }
+
+  async txCreateDoc (tx: TxCreateDoc<Doc>): Promise<void> {
+    if (this.hierarchy.isDerived(tx.objectClass, plugin.class.ActionInstance)) {
+      const instance = tx.attributes as PureActionInst
+
+      await this.storage.findAll(plugin.class.ActionInstance, { _id: instance._id })
```
haiodo, line 73: Why you need this?

server/server/src/actions/exec.ts
```diff
@@ -0,0 +1,88 @@
...
+      },
+      pop: () => stack.pop(),
+      push: (item: any) => stack.push(item),
+      create: svc.create.bind(svc),
+      update: svc.update.bind(svc),
+      subscribe: svc.subscribe.bind(svc),
+      findAll: svc.findAll.bind(svc)
+    }
+
+    if (atom._type === 'goto') {
```
haiodo, line 40: Could you have map with types and handlers?

server/server/src/actions/runtime.ts
sumboid: Yep, seems to be unnecessary

server/server/src/actions/exec.ts
sumboid: Let's update it a bit later, it doesn't seem to become more simple without changing rest of logic: handler context would be huge

haiodo (approved):

PR #293, opened 2021-08-25
Allow to mark spaces as starred for current user

haiodo:
1. Account transaction space for model operations. So it is possible to modify model objects in individual basis and hold some personalized information.
2. Starrted support for Chunter.
3. $ne operator support to allow checks for value not equal.
4. Fix Live Queries in case of update operation and previous document was not matched.


packages/core/src/storage.ts
```diff
@@ -22,6 +22,7 @@ import type { Tx } from './tx'
 export type QuerySelector<T> /* eslint-disable-line @typescript-eslint/consistent-type-definitions */ = {
   $in?: T[]
   $like?: string
+  $ne?: any
```
BykhovDenis, line 25: Why any? I think T

BykhovDenis (approved):

packages/core/src/storage.ts
haiodo: Fixed.

PR #294, opened 2021-08-25
Query fixs

BykhovDenis:
Minor query fixes & additional tests

### DIFF
--- a/packages/query/src/index.ts
+++ b/packages/query/src/index.ts
@@ -159,8 +159,8 @@ export class LiveQuery extends TxProcessor implements Storage, Queriable {
           q.result.splice(pos, 1)
           q.total--
         }
+        q.result[q.result.findIndex((p) => p._id === updatedDoc._id)] = updatedDoc
         this.sort(q, tx)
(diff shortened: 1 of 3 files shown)


packages/query/src/index.ts
```diff
@@ -153,29 +154,23 @@ export class LiveQuery extends TxProcessor implements Storage, Queriable {
...
       if (updatedDoc !== undefined) {
         const pos = q.result.indexOf(updatedDoc)
-        await this.doUpdateDoc(updatedDoc, tx)
+        this.doUpdateDoc(updatedDoc, tx)
         if (!this.match(q, updatedDoc)) {
           q.result.splice(pos, 1)
           q.total--
         }
         this.sort(q, tx)
+        q.result[q.result.findIndex(p => p._id === updatedDoc._id)] = updatedDoc
```
haiodo, line 163: I think we need to sort after document is updated and you could use pod already defined

haiodo (changes requested):

haiodo (approved):

PR #295, opened 2021-08-25
Core tests

BykhovDenis:
increased test coverage

### DIFF
--- a/packages/core/src/__tests__/memdb.test.ts
+++ b/packages/core/src/__tests__/memdb.test.ts
@@ -60,6 +60,11 @@ describe('memdb', () => {
       kind: 0
     })
     expect(first.length).toBe(1)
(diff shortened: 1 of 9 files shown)


packages/core/src/__tests__/memdb.test.ts
```diff
@@ -60,6 +60,11 @@ describe('memdb', () => {
       kind: 0
     })
     expect(first.length).toBe(1)
+    const und = await model.findAll(core.class.Class, {
+      _id: txes[1].objectId as Ref<Class<Obj>>,
+      kind: undefined
```
haiodo, line 65: did we support search for undefined?

BykhovDenis: we'll ignore undefined param

PR #302, opened 2021-08-27
Initial calendar plugin implementation

### LINKED ISSUE #259: Calendar plugin proposal
# Calendar plugin

Let's define event as:
```ts
interface Event extends Doc {
  name: string
  description: string
  startsAt: Date
  endsAt: Date
  participants: Ref<Account>[]
  eventId: string
  version: number
  isOwned: boolean
  isArchived: boolean
}
```

You may notice here `isOwned`, `eventId` and `version`, keep it in mind because later we use it to share events.

# What may happen with event
1. User is invited to event (no matter after event creation or at same time)
2. Event update
3. User is removed from event participants list

# Approaches
## Duplicate event for each participant
Every participant would have his own copy of event, that means we need to manage all copies to keep them in sync, likely we can consider only event owner can edit it.

### Handling
1. User invited to event

Adding participant mean to provide a copy of event to the participant to make him duplicate event in his calendar. Main question is how to notify user, it can be chunterbot message or automatic message from event owner, or we can even introduce "default" user space for every application so notification will be created within that space. In general, somehow user will get all data required to make event duplicate.

2. Event update

When event gets updated we need to synchronize all derived events.

3. Remove participant

If participant were removed platform has to mark his copy as archived and stop updating.

4. Add participant after removing

If user has archived event it is necessary to update it and mark as not archived.
Otherwise platform updates existing invite.

### Implementation
To implement all managing logic we need to implement event class triggers really careful. Let's take a look how they would look in first approximation.
For simplicity participant status (accepted or declined) be omitted.

First, pleast take a look at triggers and actions proposal, below is actions examples:

Let's define EventInvite like:
```ts
interface EventInvite extends Doc {
  event: Ref<Event>
  actual: boolean
  participant: Ref<Account>
}
```

#### Update event trigger
```ts
const eventUpdaterAction = Action()
  .exec((ctx: Context) => {
    const obj: Event = ctx.pop()

    // Derived event update is not our case
    if (!obj.isOwned) {
      return
    }

    const derivedEvents = await ctx.client.findAll(
      calendar.class.Event,
      {
        eventId: obj.eventId,
        modifiedBy: {$in: obj.participants},
        isOwned: false
      }
    )
    const targetEvents = derivedEvents.filter(x => x.version < obj.version)

    await Promise.all(targetEvents.map(event => ctx.client.updateDoc(
      event._class,
      event.space,
      event._id, {
        ...omitDocProps(obj),
        isOwned: false
      }
    )))
  })
  .build()
```

#### Update event participants trigger
```ts
const eventParticipantAction = Action()
  .exec((ctx: Context) => {
    const obj: Event = ctx.pop()

    if (!obj.isOwned) {
      return
    }

    const derivedEvents = await ctx.client.findAll(calendar.class.Event, {eventId: obj.eventId, isOwned: false})

    // Handle events of removed participants
    const eventsToArchive = derivedEvents.find(x => !obj.participants.has(x.modifiedBy))
    await Promise.all(eventsToArchive.map(event => ctx.client.updateDoc(
      event._class,
      event.space,
      event._id, {
        isArchived: true
      }
    )))

    // Handle invites
    const invites = await ctx.client.findAll(calendar.class.Invite, {event: obj.eventId})
    const invitesToDismiss = invites.find(x => !obj.participants.has(x.participant))

    await Promise.all(invitesToDismiss.map(invite => ctx.client.updateDoc(
      invite._class,
      invite.space,
      invite._id, {
        actual: false
      }
    )))

    // Invite participants that were not invited yet
    // For now consider user has some "default" space
    const participantsToInvite = obj.participants.filter(x => derivedEvents.every(e => e.modifiedBy !== x))
    await Promise.all(
      participantsToInvite.map(participant => ctx.client.createDoc(
        calenar.class.Invite,
        getDefaultUserSpace('calendar.app', participant),
        {
          ref: event,
          participant
          ...
          // name, description, time...
        }
      ))
    )

    // Handle users that were invited, removed and invited again
    const eventsToRestore = derivedEvents.find(x => obj.participants.has(x.modifiedBy) && x.isArchived)
    await Promise.all(eventsToArchive.map(event => ctx.client.updateDoc(
      event._class,
      event.space,
      event._id, {
        isArchived: false
      }
    )))
  })
```

#### Copy event also known as "accept invite"
For this case we can use action. We need only reference of event to make a copy:

```ts
const acceptEventInviteAction = Action()
  .exec((ctx: Context) => {
    const space: Ref<Space> = ctx.pop() // We can pass space by stack adjustment
    const invite: EventInvite = ctx.pop() // Event will be pushed after ref

    // We need to check that event owner sent invite to the action
    // owner. Otherwise there would be a security issue.
    // For now consider user has some "default" space
    if (invite.space !== getDefaultUserSpace('calendar.app', owner.id) || !invite.actual) {
      throw Error('Invalid invite')
    }

    const event = ctx.client.findAll(calendar.class.Event, {
      _id: invite.ref
    })[0]

    if (!event) {
      throw Error('Missing event')
    }

    await ctx.client.createDoc(event._class, space, {
      ...omitDocProps(event),
      isOwned: false
    })

    await ctx.client.updateDoc(invite._class, invite.space, invite._id, {
      actual: false
    })
  })
```

sumboid:
Resolves #259


models/calendar/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * @public
+ */
+@Model(calendar.class.DerivedEvent, core.class.DerivedData, DOMAIN_CALENDAR)
+export class TDerivedEvent extends TDerivedData implements DerivedEvent {
+  name!: string
+  description!: string
+  startsAt!: Date
```
haiodo, line 53: same here

models/core/src/derived.ts
```diff
@@ -41,7 +41,7 @@ const DOMAIN_TITLE = 'title' as Domain
 /**
  * @public
  */
-@Model(core.class.DerivedData, core.class.Doc, DOMAIN_MODEL)
+@Model(core.class.DerivedData, core.class.Doc)
```
haiodo, line 44: Why DOMAIN_MODEL is dropped?

packages/core/src/derived/processor.ts
```diff
@@ -206,11 +206,12 @@ export class DerivedDataProcessor extends TxProcessor {
 
     for (const d of descriptors) {
       const results = (await this.applyMapper(d, tx)) ?? (await this.applyRule(d, tx, doc))
-      const oldData = await this.storage.findAll(core.class.DerivedData, {
+      const oldData = await this.storage.findAll(d.targetClass, {
```
haiodo, line 209: Please rebase, @BykhovDenis already fixed this issue.

haiodo (changes requested):

packages/core/src/query.ts
```diff
@@ -34,7 +35,10 @@ export function findProperty (objects: Doc[], propertyKey: string, value: any):
   const result: Doc[] = []
   for (const object of objects) {
     const val = (object as any)[propertyKey]
-    if (val === value || nestedDotQueryCheck(propertyKey, object, value)) {
+    const isEqual = Array.isArray(value)
```
haiodo, line 38: deepEqual should check for array I suppose, so could we left only deepEqual?

plugins/calendar-assets/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
```
haiodo, line 1: Could you please reuse minimal .eslintrc from any existing ui plugin.

plugins/calendar-assets/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
haiodo, line 1: Please use generic ts config from platform-rig

plugins/calendar/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
```
haiodo, line 1: Please use minimal .eslitrc

plugins/calendar/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+import { plugin } from '@anticrm/platform'
+import type { Asset, IntlString, Resource } from '@anticrm/status'
+
+
+export interface Calendar extends Space {}
+
+export interface Event extends Doc {
+  name: string
+  description: string
+  startsAt: Date
```
haiodo, line 29: Please use number instead of Date.

plugins/calendar/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  endsAt: Date
+  participants: Array<Ref<Account>>
```
haiodo, line 31: Please use eslint config as in `@anticrm/core` in this case it will have eslint rule configured to use Ref<Account>[]

sumboid: For some reason eslintrc update hasn't affected this line

haiodo: Will check

haiodo (approved):

PR #304, opened 2021-08-30
Chunter notifications

BykhovDenis:


models/chunter/src/index.ts
```diff
@@ -129,6 +142,46 @@ export function createModel (builder: Builder): void {
     direct: false
   })
 
+  registerMapper(chunter.mapper.CommentNotificationMappes, {
```
haiodo, line 145: It should be called from server, since model code is executed once outside of server process to create model.

plugins/chunter-impl/src/components/Message.svelte
```diff
@@ -49,6 +53,20 @@
     return (await client.findAll(core.class.Account, { _id: userId }))[0]
   }
 
+  let query: QueryUpdater<Notification> | undefined
+
+  query = client.query(
```
haiodo, line 58: I think we need to try not call any query per individual messages, if we had it already we need to try get rid of it.

plugins/chunter-impl/src/components/Message.svelte
```diff
@@ -36,6 +39,7 @@
 
   export let message: MessageData
   export let thread: boolean = false
+  let notifications: Array<Notification> = []
```
haiodo, line 42: Why it could be multiple of notifications per individual message?

plugins/notification-impl/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export { notificationPlugin }
+
+export default async (): Promise<NotificationService> => {
+  const coreP = await getPlugin(corePlugin.id)
+  const client: Client = await coreP.getClient()
+  const accountId = client.accountId()
+
+  async function subscribe<T extends Doc> (_class: Ref<Class<T>>, space: Ref<Space>, objectId: Ref<T>): Promise<void> {
+    console.log('SUBSCRIBE CALL')
```
haiodo, line 30: Do we need it?

PR #325, opened 2021-09-07
Extract ENV variables into separate .env file

keepitsimple:
It's recommended do not keep sensitive info like SERVER_SECRET in the files stored in the GIT repo.

Standard approach is to use `.env` file which is ignored in GIT.
The `.env.sample` file is provided as example


haiodo:
It is docker container default, anyone how would like to deploy container will pass own values, I've do not see why they should be outside of docker container definition file.

keepitsimple:
@haiodo multiple reasons:
1. It's a bad practice; early or later somebody will commit real secrets into this file and it will appear in a public git repo.  
2. "will pass own values" - yep and the recommended way to do it via external ENV variables passed outside of the container.

PR #360, opened 2021-09-15
Document selection support

haiodo:
new Workbench API to selectDocument on aside.


packages/ui/src/components/Table.svelte
```diff
@@ -73,9 +73,15 @@
...
         {#each data as doc (doc._id)}
-          <tr class="tr-body" on:click={() => dispatch('rowClick', { id: doc._id })}>
+          <tr class="tr-body" on:click={() => dispatch('rowClick', doc)}>
             {#each docToRow(doc) as cell}
-              <td><svelte:component this={cell.component} {...cell.props} /></td>
+              <td
+                ><svelte:component
+                  this={cell.component}
+                  {...cell.props}
+                  on:click={() => dispatch('rowClick', doc)}
```
BykhovDenis, line 82: Why we need both events on a row and a celll?

haiodo: Good catch, we didn't.

plugins/task-assets/lang/en.json
```diff
@@ -29,6 +29,7 @@
     "InProgress": "In progress",
     "Closed": "Closed",
     "Favorite": "Favorite",
-    "MyTasks": "My tasks"
+    "MyTasks": "My tasks",
+    "PleaseSelectSpace": "Space selection is required."
```
BykhovDenis, line 33: Maybe use a "project" for tasks? I'm not sure because we have specials

plugins/workbench-impl/src/png.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*.png'
```
BykhovDenis, line 1: Why we need it?

BykhovDenis (approved):

plugins/chunter-impl/src/components/CreateDirectMessage.svelte
```diff
@@ -105,7 +105,9 @@
   }
 
   afterUpdate(() => {
-    notificationClient.initScroll(div, notifications?.lastRead ?? 0, false)
+    if (div) {
```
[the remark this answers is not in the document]
haiodo: In my env, it could be null

plugins/workbench-impl/src/png.d.ts
haiodo: svelte-check is not passed if png files are imported without this kind of definition.

plugins/task-assets/lang/en.json
haiodo: I think we need to do something more interesting, like showing some dashboard etc. So let it be for now.

sumboid (approved):

PR #365, opened 2021-09-20
SSL/TLS support

haiodo:
Enable SSL/TLS support for both development and production HR build.


products/hr-build/readme.md
```diff
@@ -1,22 +1,38 @@
...
+- SECURITY_CERT_FILE - Certificate file in PEM format.
+- SECURITY_KEY_FILE - Private key file in PEM format.
+- SECURITY_CA_FILE - Certificate Chain in PEM format.
+
+## Starting server
+
+Running server as background process could be achived with.
+
+```bash
+notail node ./server.js &
```
sumboid, line 37: I know that `nohup` can help to detach process but have never heard about `notail`

BykhovDenis (approved):

sumboid (commented):
What a point to use SSL on localhost? Isn't it easier to use it only when necessary, for specific builds for instance?

sumboid:
FYI: just checked on linux env, works fine

haiodo:
Some stuff like auto insert of email/passwords is working differently, not sure about more stuff, so I think it is better to find all such places in dev.

sumboid (approved):

PR #370, opened 2021-09-21
New DatePicker, add Link

SasLord:


packages/ui/src/components/popups/DatePickerPopup.svelte
```diff
@@ -0,0 +1,160 @@
...
+    <button
+      class="arrow"
+      on:click|preventDefault={() => {
+        view.setMonth(view.getMonth() + 1)
+        view = view
+      }}><Forward size={16} /></button
+    >
+  </div>
+  <div class="calendar">
+    <div class="flex-center caption">Mo</div>
```
haiodo, line 80: we need to use internationalisation routines to obtain all of this information instead of predefined ones.

haiodo: @SasLord let's create an issue to handle it as next step.

haiodo (approved):

PR #390, opened 2021-09-28
Separate DD State Space

haiodo:
Separate DD State Space


packages/core/src/component.ts
```diff
@@ -51,7 +51,8 @@ export default component(ComponentCore, {
   },
   space: {
     Tx: '' as Ref<Space>, // A special space for all transactions
-    Model: '' as Ref<Space> // A special space for all model transactions.
+    Model: '' as Ref<Space>, // A special space for all model transactions.
+    DerivedData: '' as Ref<Space> // A special space for all model transactions.
```
BykhovDenis, line 55: Can we fix or remove comments?)

haiodo: Fixed.

PR #394, opened 2021-09-28
Notification fix

BykhovDenis:


plugins/chunter-impl/src/components/ChannelView.svelte
```diff
@@ -34,7 +34,7 @@
 
   let messages: Message[] = []
 
-  async function addMessage (message: string): Promise<void> {
+  async function addMessage (message: string, notifications: SpaceNotifications | undefined): Promise<void> {
```
haiodo, line 37:
```suggestion
  async function addMessage (message: string, notifications?: SpaceNotifications): Promise<void> {
```

plugins/chunter-impl/src/components/CreateDirectMessage.svelte
```diff
@@ -80,7 +80,7 @@
     }
   })
 
-  async function addMessage (message: string): Promise<void> {
+  async function addMessage (message: string, notifications: SpaceNotifications | undefined): Promise<void> {
```
haiodo, line 83:
```suggestion
  async function addMessage (message: string, notifications?: SpaceNotifications): Promise<void> {
```

plugins/chunter-impl/src/components/ThreadsView.svelte
```diff
@@ -61,7 +61,7 @@
     selectDocument(undefined)
   }
 
-  async function addMessage (text: string): Promise<void> {
+  async function addMessage (text: string, notification: SpaceNotifications | undefined): Promise<void> {
```
haiodo, line 64:
```suggestion
  async function addMessage (text: string, notification?: SpaceNotifications): Promise<void> {
```

plugins/notification-mappers/src/index.ts
```diff
@@ -98,35 +98,116 @@ function createSpaceInfo<T extends Space> (
...
+    result.push(spaceNotification)
+  }
+}
+
+async function createDocSpaceInfoHandler (ctx: TxCreateDoc<Doc>, options: MappingOptions): Promise<SpaceInfo[]> {
+  if (options.hierarchy.isDerived(ctx.objectClass, core.class.Space)) {
+    return [createSpaceInfo(ctx as TxCreateDoc<Space>, options.descriptor._id)]
+  }
+  await tryUpdateSpaceInfo(ctx, options)
+  return []
```
haiodo, line 118: Why we not return updated value, does DD processor will diff and remove existing DD? Should we return updated value as well? I've created ticked #395 to disable processor for doing anything if mapper is designed for it,

BykhovDenis: Because SpaceInfo is derived data from space, not from object in.

haiodo (approved):

Issue #398, opened 2021-09-29
Chunter: Implement editing of Messages/Comments by author.


haiodo:
Chunter: Implement editing of Messages/Comments by author.

PR #410, opened 2021-09-30
Allow to edit chunter messages/comments

haiodo:
Allow to edit chunter messages/comments

Closes: #398

### DIFF
--- a/plugins/chunter-impl/src/components/Message.svelte
+++ b/plugins/chunter-impl/src/components/Message.svelte
@@ -17,10 +17,17 @@
   import chunter from '@anticrm/chunter'
   import core, { Account, Class, Doc, parseFullRef, Ref, Timestamp } from '@anticrm/core'
   import type { SpaceNotifications } from '@anticrm/notification'
-  import type { MessageNode } from '@anticrm/text'
-  import { parseMessage } from '@anticrm/text'
-  import type { ItemRefefence } from '@anticrm/ui'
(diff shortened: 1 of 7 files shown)


plugins/chunter-impl/src/components/Message.svelte
```diff
@@ -135,43 +156,103 @@
...
-          <span>
-            <DateTime value={message.modifiedOn} timeOnly={isToday(message.modifiedOn)} />
-          </span>
-        </div>
-        {#if !thread}
+      {#if !editMode}
+        <div class="header">
+          <div>
+            {#if user}
+              {user?.name ?? ''}
```
BykhovDenis, line 163: Why we need `user?.` in `#if user`

plugins/chunter-impl/src/components/Message.svelte
```diff
@@ -53,6 +62,11 @@
 
   const client = getClient()
 
+  $: currentId = client.accountId()
```
BykhovDenis, line 65: Why not const?

plugins/chunter-impl/src/components/Message.svelte
haiodo: Not required. Thanks.

plugins/chunter-impl/src/components/Message.svelte
haiodo: Fixed.

BykhovDenis (approved):

PR #419, opened 2021-10-04
Event token for set metadata

### LINKED ISSUE #414: Use token event for set metadata

BykhovDenis:
close #414

### DIFF
--- a/dev/prod/src/main.ts
+++ b/dev/prod/src/main.ts
@@ -36,13 +36,13 @@ setMetadata(meetingPlugin.metadata.ClientUrl, `${meetingHost}:${meetingPort}`)
 setMetadata(pluginCore.metadata.ClientUrl, clientUri)
 
 addEventListener('Token', async (event, data) => {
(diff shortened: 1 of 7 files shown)


dev/prod/src/main.ts
```diff
@@ -36,13 +36,13 @@ setMetadata(meetingPlugin.metadata.ClientUrl, `${meetingHost}:${meetingPort}`)
 setMetadata(pluginCore.metadata.ClientUrl, clientUri)
 
 addEventListener('Token', async (event, data) => {
+  setMetadata(pluginCore.metadata.Token, data)
```
haiodo, line 39: What about AccountId? It is also required by `plugin-core`?

BykhovDenis: I think it is not needed and is outdated, I could not find a place where it would be used.

haiodo: Ok, we need to remove it in this case.

haiodo (approved):

PR #444, opened 2021-10-07
New popups

SasLord:


packages/ui/src/components/Card.svelte
```diff
@@ -0,0 +1,163 @@
...
+  {#if $$slots.pool}
+    <div class="flex-col pool" class:shrink={$$slots.contacts}>
+      <div class="separator" />
+      <slot name="pool" />
+    </div>
+  {/if}
+  <div class="footer">
+    <Button
+      disabled={!canSave}
+      label={'Ok'}
```
haiodo, line 56: Please use ui.string.Ok etc.

packages/ui/src/components/Card.svelte
```diff
@@ -0,0 +1,163 @@
...
+      size={'small'}
+      transparent
+      primary
+      on:click={() => {
+        okAction()
+        dispatch('close')
+      }}
+    />
+    <Button
+      label={'Cancel'}
```
haiodo, line 66: Same here.

packages/ui/src/components/Component.svelte
```diff
@@ -22,15 +23,34 @@
 
   export let is: AnyComponent
   export let props: any = {}
+  export let scrollable: boolean = false
 
   $: component = getResource(is)
+
+  const dispatch = createEventDispatcher()
+  afterUpdate(() => {
+    dispatch('update', Date.now())
```
haiodo, line 32: still not sure why we need this trick.

plugins/recruiting-impl/src/components/candidates/EditCandidate copy.svelte
```diff
@@ -0,0 +1,105 @@
+<!--
```
haiodo, line 1: Why we need this file?

plugins/recruiting-impl/src/components/vacancies/VacancyEditor copy.svelte
```diff
@@ -0,0 +1,123 @@
+<!--
```
haiodo, line 1: Why we need this file?

haiodo:
@SasLord AsideDocument is not modified, do you plan to deligate it to some of us?

PR #452, opened 2021-10-18
Attachment preview

### LINKED ISSUE #393: Attachments: Preview and img compression

BykhovDenis:
close #393


server/s3/src/index.ts
```diff
@@ -21,6 +21,14 @@ import https from 'https'
  */
 export type Body = Buffer | Uint8Array | Blob
 
+/**
+ * @public
+ */
+export interface File {
+  body: Buffer
+  type: string | undefined
```
haiodo, line 29:
```suggestion
  type?: string
```

server/s3/src/index.ts
```diff
@@ -15,18 +15,28 @@
...
+  type?: string
+}
+
 /**
  * @public
  */
 export class S3Storage {
   private readonly client: S3
   private readonly bucket: string
+  private readonly imageCache: Map<string, File> = new Map<string, File>()
```
haiodo, line 39: why we need cache in memory?

PR #524, opened 2021-11-01
Table virtualization

### LINKED ISSUE #484: Table virtualization

sumboid:
Resolves #484

### DIFF
--- a/packages/ui/src/components/Table.svelte
+++ b/packages/ui/src/components/Table.svelte
@@ -17,10 +17,11 @@
   import type { IntlString, UIComponent } from '@anticrm/status'
 
   import Label from './Label.svelte'
-  import ScrollBox from './ScrollBox.svelte'
+  import VirtualList, { sty } from './VirtualList.svelte'
 
   interface Cell {
     component: UIComponent
+    width?: number
     props?: Object
   }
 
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Table.svelte
```diff
@@ -57,67 +60,86 @@
         {}
       )
     }))
+
+  const widthSty = (f: { width?: number }) =>
+    f.width !== undefined ? `min-width: ${f.width}px; width: ${f.width}px` : ''
+
+  let rerender = false
+  export function reset () {
```
haiodo, line 68: Could we avoid this kind of dirty tricks? It is unclear about it should be called outside for proper functionality.

Please check for data field updates.

plugins/recruiting-impl/src/components/candidates/CandidateTable.svelte
```diff
@@ -38,10 +38,12 @@
     }
   ]
 
+  let table: any
   const client = getClient()
   let data: Candidate[] = []
   let lq: QueryUpdater<Candidate> | undefined
   $: if (currentSpace !== prevSpace) {
+    table?.reset()
```
haiodo, line 46: No way we should do so.

haiodo (approved):

PR #539, opened 2021-11-05
Candidates

BykhovDenis:


models/dev/src/demoAttachment.ts
```diff
@@ -27,6 +27,7 @@ export async function createAttachment (
     {
       objectId: objectId,
       objectClass: objcetClass,
+      attachTo: getFullRef(objectId, objcetClass),
```
haiodo, line 30: May be we should have only attachTo instead of objectId, objectClass?

packages/ui/src/components/Table.svelte
```diff
@@ -71,6 +72,19 @@
     list?.reset()
   }
   let checked: boolean = false
+
+  function asComponent (comp: UIComponent | AnyComponent): AnyComponent {
+    return comp as AnyComponent
+  }
+
+  function getValue (doc: any, key: string): any {
```
haiodo, line 80: Could you try to use '@anticrm/core' getNestedValue() method for same, it also could query for arrays.

packages/ui/src/components/Table.svelte
```diff
@@ -106,7 +122,11 @@
                     <div class="menuBtn"><MoreV size={16} /></div>
                   </div>
                 {/if}
-                <svelte:component this={cell.component} {...cell.props} />
+                {#if typeof cell.component === 'string'}
+                  <Component is={asComponent(cell.component)} props={cell.props} />
```
haiodo, line 126: Let's check if this construction is used in any other place and if so, let's move this logic inside Component itself.

packages/ui/src/index.ts
```diff
@@ -74,8 +74,10 @@ export { default as Panel } from './components/Panel.svelte'
 export { default as PopupWrap } from './components/popups/PopupWrap.svelte'
 export { default as YesNo } from './components/YesNo.svelte'
 export { default as YesNoCard } from './components/YesNoCard.svelte'
+export { default as Header } from './components/Header.svelte'
```
haiodo, line 77: I propose to think about better name based on it's logic.

plugins/chunter-impl/src/components/CommentsTableCell.svelte
```diff
@@ -0,0 +1,38 @@
...
+    if (items.length > 0) {
+      e.stopPropagation()
+      showPopup(CommentsPopup, { comments: items }, e.target as HTMLElement)
+    }
+  }
+</script>
+
+<div class="flex-row-center" on:click={open}>
+  <Chat />
+  {comments.length}
```
haiodo, line 37: Do we need it here?

plugins/contact/src/index.ts
```diff
@@ -26,12 +26,9 @@ export interface Person extends Doc {
   lastName: string
   avatar?: string
   bio?: string
-  phone?: string
```
haiodo, line 29: Why they are removed?

plugins/contact/src/index.ts
```diff
@@ -26,12 +26,9 @@ export interface Person extends Doc {
...
   address: {
     country?: string
     city?: string
-    street?: string
```
haiodo, line 33: why removed?

haiodo (approved):

PR #554, opened 2021-11-08
Fix 541

### LINKED ISSUE #541: It's possible to assign candidate twice to the vacancy

BykhovDenis:
Close #541

### DIFF
--- a/plugins/recruiting-impl/src/components/candidates/CreateApplication.svelte
+++ b/plugins/recruiting-impl/src/components/candidates/CreateApplication.svelte
@@ -33,12 +33,16 @@
   const dispatch = createEventDispatcher()
 
   let applicants: Applicant[] = []
+  let applicantsLoaded = false
   let applicantsQ: QueryUpdater<Applicant> | undefined
   $: applicantsQ = client.query(
     applicantsQ,
     recruiting.class.Applicant,
     { _id: { $in: candidate.applicants } },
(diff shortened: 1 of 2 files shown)


haiodo (approved):

plugins/recruiting-impl/src/components/candidates/CreateApplication.svelte
```diff
@@ -112,11 +110,20 @@
 
   let canSave: boolean = false
   $: canSave = !!(recruiter && candidate && stateID)
+
+  function getAvaibleVacancies (applicants: Applicant[]): DropdownItem[] {
```
sumboid, line 114: typo, and why not to just define `availableVacancies` and update it by `$`? Also I think that vacancies should be empty until we receive applicants

BykhovDenis: Typo fixed. Because we don't have `not in` query and query updater not update callback func

sumboid: @BykhovDenis I mean something like this:
```ts
let availableVacancies = []
$: if (applicantsLoaded) {
  availableVacancies = vacancyItems
    .filter((item) => !applicants
      .map((app) => app.space)
      .includes(item.id as Ref<Space>))
}
```

sumboid (approved):

PR #567, opened 2021-11-11
Activity notifications

### LINKED ISSUE #560: Notification: Add support of activity views
Notification: Add support of activity views

BykhovDenis:
close #560


plugins/task-impl/src/components/KanbanCard.svelte
```diff
@@ -46,10 +46,12 @@
 
   function isNew (card: Task, spaceLastViews: SpaceLastViews | undefined): boolean {
     if (spaceLastViews === undefined) return false
```
haiodo, line 48: I think we need to extract code like this into some utility place

haiodo (approved):

